PR #1, opened 2017-03-11
Acq/File Info extensions

jrs65:
This should be mostly finished now.

There was some issue getting one test to pass. But i'm not sure why that test makes sense so I've commented it out. We can chat about it on Monday?


alpenhorn/acquisition.py
```diff
@@ -25,6 +32,114 @@ class AcqType(base_model):
...
+        reg_acq_types = cls._registered_acq_types.keys()
+
+        # Find any missing types
+        missing_acq_types = set(db_acq_types) - set(reg_acq_types)
+
+        if len(missing_acq_types):
+            raise RuntimeError('AcqTypes %s have no registered handler.' %
+                               repr(missing_acq_types))
+
+    def is_type(self, acqname, node):
```
cubranic: This method doesn't do much, why not do the check in `detect` where it calls this method?

cubranic: Ah, is it intended as a public method that could be called in situations other than import? I only see the call from `detect` right now, but you may have other uses in mind already.

jrs65: Yeah, exactly. It's intended to be a public method that can be called by other code to tell if an acquisition is of the correct type. It also only passes on the root directory of the node and not the full StorageNode object so that it can't be modified.

alpenhorn/auto_import.py
```diff
@@ -67,54 +69,56 @@ def _import_file(node, root, acq_name, file_name):
...
         md5sum = md5sum_file(fullpath, cmd_line=False)
         size_b = os.path.getsize(fullpath)
+
         done = False
         while not done:
             try:
-                file = ac.ArchiveFile.create(acq=acq, type=ftype, name=file_name,
-                                             size_b=size_b, md5sum=md5sum)
+                with db.database_proxy.atomic():
+                    file_ = ac.ArchiveFile.create(acq=acq, type=ftype, name=file_name,
```
cubranic: Should `file_` be `file`? See why below at line 117.

alpenhorn/auto_import.py
```diff
@@ -67,54 +69,56 @@ def _import_file(node, root, acq_name, file_name):
...
+
                 # di.connect_database(True)
         log.info("File \"%s/%s\" added to DB." % (acq_name, file_name))
 
     # Register the copy of the file here on the collection server, if (1) it does
     # not exist, or (2) it does exist but has been labelled as corrupt. If (2),
     # check again.
-    if not file.copies.where(ar.ArchiveFileCopy.node == node).count():
-        copy = ar.ArchiveFileCopy.create(file=file, node=node, has_file='Y',
+    if not file_.copies.where(ar.ArchiveFileCopy.node == node).count():
```
cubranic: `_file` is only set if the file is new, but this will fail if ArchiveFile does exist in the DB but does not have a corresponding ArchiveFileCopy on this node. We also don't check for the second case described as the comment above, if ArchiveFileCopy exists but has been marked as corrupt.

alpenhorn/acquisition.py
```diff
@@ -25,6 +32,114 @@ class AcqType(base_model):
...
+        def _resolve(x):
+            if isinstance(x, str):
+                return x
+            elif isinstance(x, type) and issubclass(x, FileInfoBase):
+                return x.file_type
+            else:
+                raise RuntimeError("FileType %s not understood." % repr(x))
+
+        # Resolve the file_types list of the AcqInfo to the names
+        file_type_names = [_resolve(x) for x in self.acq_info._file_types]
```
cubranic: Why can `_file_types` be either strings or instances of FileInfoBase?

alpenhorn/acquisition.py
```diff
@@ -25,6 +32,114 @@ class AcqType(base_model):
...
+            cls.get(name=acq_info._acq_type)
+        except pw.DoesNotExist:
+            log.info("Create AcqType entry for \"%s\"" % acq_info._acq_type)
+            cls.create(name=acq_info._acq_type)
+
+        # Add to registry
+        cls._registered_acq_types[acq_info._acq_type] = acq_info
+
+    @classmethod
+    def check_registration(cls):
```
cubranic: Call `check_registrations` (plural)? Or also include "handler" and/or "type"

alpenhorn/acquisition.py
```diff
@@ -63,6 +178,99 @@ class FileType(base_model):
...
+            The acquisition the file is in.
+        node : StorageNode
+            The node we are importing from. Needed so we can inspect the actual
+            acquisition.
+
+        Returns
+        -------
+        is_type : boolean
+        """
+        return self.file_info._is_type(filename, path.join(node.root, acq.name))
```
cubranic: I'm still confused why acq- and file-type handlers need to know about the node root in relation to the acquisition name or file path? Isn't it potentially different on every node? It seems to me like the all of the paths should be root-relative, and then the code that operates on actual physical files on the disk simply prefixes it with the root of the node where it's operating.

I know this specific methods usually run only on a single node that does the importing (e.g., at DRAO), but don't we want to allow in principle multiple import locations, for instance if some expensive post-processing or analysis is done elsewhere?

alpenhorn/auto_import.py
jrs65: Hmmmm. I meant to change them all to `file_`. I changed it so as not to mask the builtin type. I'll go through and fix that so it's consistent.

alpenhorn/acquisition.py
jrs65: It's not so that the handler can perform processing based on the node root name, it's so the handler can construct the full path to the file it's trying to import so it can open the file and look at it. The acq/file handlers have no idea which node is currently being processed, so if you only gave the code the `(filename, acq.name)` it couldn't open the file.

I don't think this restricts you to a single node for importing at all.

alpenhorn/acquisition.py
jrs65: `_file_types` has the option of being a string such that it can easily be set from a config file. In theory the resolve step could be done as the configuration is set, but I figured it was easy enough to do it here. As for why it can also take an instance, that seems like it could be useful and I figured why not?

alpenhorn/acquisition.py
cubranic: Ah, right, I didn't have it straight in my mind when I was thinking about the `set_info` methods. For some reason I thought the node root leaks into the file/archive info tables.

alpenhorn/auto_import.py
cubranic: And the check for the corruption?

cubranic:
OK.

alpenhorn/auto_import.py
jrs65: Yeah. It seems like the comment is wrong. Maybe some code go removed in the past. I think what we want to do is if the file was marked as corrupt to change it to `has_file='M'` to force the check of its `md5`.

jrs65: I'll send in a new commit in a few minutes.

jrs65:
I've added a new commit, with some logic to cover that corruption case. Basically, if the file has been on the node before it ensures there is one ArchiveFileCopy marked with `has_file='M'` (to force a verification of its md5), and all others get marked with `has_file='N'`.

PR #5, opened 2017-03-21
Restore the old watchdog functionality into alpenhorn.auto_import

cubranic:


cubranic:
The check passes on my machine, but try it on yours too. it could be Travis needs a longer pause for the watchdog to get notified about the new files, or maybe there is something about me running on a Mac laptop. Any ideas how to do this more reliably?

tests/test_import.py
```diff
@@ -218,3 +222,50 @@ def test_import_corrupted(fixtures):
...
+
+    # this file is outside any acqs and should be ignored
+    g = tmpdir.join('some_file.log')
+    g.write('Where is my acq?!')
+
+    # nothing happens until the observer is running
+    assert file_copy_count('new_file.log') == 0
+    assert file_copy_count('some_file.log') == 0
+
+    obs.start()
```
jrs65: It's not obvious to me that the watchdog should notice filesystem events before `obs.start` is called. In that case you wouldn't expect the test to pass (maybe MacOS has slightly different behaviour which means it does). Can we shift the test to, create the files before the watchdog starts, test that nothing happens, and then rm them, start the watchdog, recreate them, and test that they do get imported.

cubranic: That was it! Passing now...

tests/test_import.py
```diff
@@ -218,3 +222,50 @@ def test_import_corrupted(fixtures):
...
+
+    f = acq_dir.join('new_file.log')
+    f.write('')
+
+    # this file is outside any acqs and should be ignored
+    g = tmpdir.join('some_file.log')
+    g.write('Where is my acq?!')
+
+    # nothing happens until the observer is running
+    assert file_copy_count('new_file.log') == 0
```
jrs65: These assertions seem a little fragile as they rely on the latency of the watchdog making the import call being longer than the time it takes from writing the file to getting to this point. Similarly if the latency was longer than expected (> 1s) the tests below might fail.

cubranic: In retrospect, we don't want to test `watchguard`, just our event handler. I'll change the tests accordingly.

PR #6, opened 2017-04-05
Allow nested acquisition directory and file structure (fixes #4)

### LINKED ISSUE #4: Handle acquisitions as nested directories of the root and files in acq subdirs
Desired use case from FRB:

Acqusitions (::= EVT_DIR):
```
<CHIME_FRB_ROOT>/<yyyy>/<mm>/<dd>/<frb|rfi>_<event_id>_<UTCdatetime>/(event details)
```

Data files:
```
EVT_DIR/summary.tx
EVT_DIR/intensity_data/beam_<<beam_no>>/  (repeat for multiple beams)
EVT_DIR/intensity_data/beam_<<beam_no>>/raw/
EVT_DIR/intensity_data/beam_<<beam_no>>/proc/
EVT_DIR/baseband_data/raw/
EVT_DIR/baseband_data/proc/
EVT_DIR/housekeeping_data/
```

cubranic:
* Importing a file asks each of the registered acquisition types if the dirname belongs to it. If none can handle it, the rightmost segment of the directory is removed and the process started again.
* First match wins, and the rest of the path to the file becomes the file name.
* The file name is tried on each of the acquisition's file types, and in case of no matches the file is skipped -- we don't try to repeat the loop on the subset of the file name's path segments as we do with acquisitions.
* The `auto_import.import_file` now has one fewer argument, since has to decide where the file's path gets broken into the acquisition vs the file name.

**Please note:** a number of TODOs are my questions over some implementation details. Let me know what you think.

### DIFF
--- a/alpenhorn/acquisition.py
+++ b/alpenhorn/acquisition.py
@@ -105,14 +105,17 @@ class AcqType(base_model):
             The node we are importing from. Needed so we can inspect the actual
             acquisition.
         """
+        # TODO: document return type
 
-        # Iterate over all known acquisition types to try and find one that matches
-        # the directory being processed
-        for acq_type in cls.select():
-
-            if acq_type.is_type(acqname, node):
-                return acq_type
-
+        # Iterate over all known acquisition types to try and find one that
+        # matches the directory being processed. If nothing is found, repeat
+        # the process with the parent directory of acqname, until we run out of
+        # directory segments
+        while acqname != '':
+            for acq_type in cls.select():
+                if acq_type.is_type(acqname, node):
+                    return acq_type, acqname
+            acqname = path.dirname(acqname)
         return None
 
     @property
--- a/alpenhorn/acquisition.py
+++ b/alpenhorn/acquisition.py
@@ -104,11 +104,23 @@ class AcqType(base_model):
         node : StorageNode
             The node we are importing from. Needed so we can inspect the actual
             acquisition.
+
+        Returns
+        -------
+        (AcqType, str) or None
+            A tuple of acquisition type and name or None if one could not be found
+
+        Notes
+        -----
+
+        The returned acquisition name is either equal to the parameter
+        `acqname` or the closest ancestor directory which had a matching
+        `AcqType`.
+
         """
-        # TODO: document return type
 
         # Iterate over all known acquisition types to try and find one that
-        # matches the directory being processed. If nothing is found, repeat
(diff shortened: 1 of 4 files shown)


jrs65:
Davor. Sorry for the delay. Been swamped at site. These changes look great to me.

alpenhorn/auto_import.py
```diff
@@ -175,12 +184,10 @@ def add_acq(name, node, comment=""):
         raise AlreadyExists("Acquisition \"%s\" already exists in DB." %
                             name)
 
-    # Find an acquisition type that can handle this acq
-    acq_type = ac.AcqType.detect(name, node)
-
-    if acq_type is None:
-        log.debug("No handler available to process \"%s\"" % name)
-        return
+    # TODO: should we still check?
```
jrs65: I think we don't need to check any more at this point. Other than as a validation of arguments (in which case we should raise `ValueErrror`).

alpenhorn/auto_import.py
```diff
@@ -44,17 +44,21 @@ def _import_file(node, root, acq_name, file_name):
...
+    dir_name = os.path.dirname(file_path)
+    base_name = os.path.basename(file_path)
+    if os.path.isfile(os.path.join(root, dir_name, ".%s.lock" % base_name)):
         log.debug("Skipping \"%s\", which is locked by ch_master.py." % fullpath)
         return
 
     # Check if we can handle this acquisition, and skip if we can't
-    if ac.AcqType.detect(acq_name, node) is None:
-        log.info("Skipping non-acquisition path %s." % acq_name)
+    # TODO: detect on the full file_path or just dir_name?
```
jrs65: I think the full file path is the most natural thing here, i.e. the purpose is to detect what type of acq this is given the file we are importing.

alpenhorn/acquisition.py
```diff
@@ -105,14 +105,17 @@ def detect(cls, acqname, node):
...
+        # Iterate over all known acquisition types to try and find one that
+        # matches the directory being processed. If nothing is found, repeat
+        # the process with the parent directory of acqname, until we run out of
+        # directory segments
+        while acqname != '':
+            for acq_type in cls.select():
+                if acq_type.is_type(acqname, node):
+                    return acq_type, acqname
+            acqname = path.dirname(acqname)
         return None
```
jrs65: I think this is the right think to return (the other option being `(None, None)`.

PR #10, opened 2017-04-24
Configuration loading support

jrs65:
I've added support for loading configuration files. The paths it checks are basically the ones recommended in the XDG standards, but that means they are a little linux specific (especially the /etc/xdg/ one). But I think that's alright.


jrs65:
I'm going to merge this as there seem to be no objections, and I think it's pretty uncontroversial.

PR #11, opened 2017-04-24
Python 3 port

jrs65:
A minimal porting of alpenhorn to support Python 3. Done using python-future (though doesn't need its compatibility library).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,10 @@
 language: python
+
+python:
+    - "2.7"
+    - "3.5"
+    - "3.6"
+
 install:
     - pip install .
     - pip install -r requirements.txt
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: python
 
 python:
-    - "2.7"
+#    - "2.7"
     - "3.5"
-    - "3.6"
+    #- "3.6"
 
 install:
     - pip install .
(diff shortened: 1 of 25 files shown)


cubranic:
Nice! I’m slightly miffed that we’re stuck with 2.7 for the FRB pipeline. 

Are we trying to be friendly to both 2.7 and 3?

Davor

jrs65:
Yeah. I figured we may as well try and explicitly support Python 3. At the moment for cosmology we're on Python 2.7, but I have a vague plan to try and slowly port everything onto Python 3. The writing is on the wall for 2.7 (new IPython releases will no longer support it).

jrs65:
Also, I think the biggest issue stopping these tests passing is a failure in some code you introduced in `test_client.py`. Any idea what is going on?

cubranic:
Not the slightest, I don’t have any experience with Python 3. My guess would be that the mocking works slightly differently. Or maybe some regex switch, I don’t use much else in those tests.

jrs65:
Ah no. I don't mean that failure (there's one that seems to be a version conflict installing on Python 3.5). I mean the failure for Python 2.7 on the master branch. The same failure also occurs for Python 2.7 on this branch, and Python 3.6 on this branch.

cubranic:
Hm, weird. Passes on my machine. Probably some platform-dependent details of how the order in which directory contents are reported in that “unknown files” output. Let me test that...

cubranic:
Yep. That's what it was. Sigh.

jrs65:
Similarly this one I think is pretty uncontroversial.

PR #12, opened 2017-04-24
Debugging

cubranic:


jrs65:
Maybe we should just patch the`import_files` routine to sort the file lists before output?

cubranic:
Not a bad idea. The output on Linux (or at least in Travis) was logically backward in some places.

jrs65:
Seems good now! I'll merge it in.

PR #13, opened 2017-04-26
Alpenhorn service and extension support

jrs65:
This is a moderately large set of changes that implement support for
loading extensions into alpenhorn from the given configuration, and
has the first steps to getting `alpenhornd` running.

In addition to the headline changes it has a number of smaller additions:
  - A new, working, implementation of ENUM for all database types.
  - Removed SqliteEnum hacks from tests as Sqlite should be supported
    natively now.
  - Rewrote database connection code to work with the config.
  - Added an `init` command to `alpenhorn` for creating the required tables.
  - Added Python versions back to Travis testing.
  - Added script entry points back into `setup.py`
  - Removed some exception classes mistakenly brought in from the layout db.
  - Added a custom database connection class that is a hybrid of Proxy and
    RetryOperationalError.

### DIFF
--- a/alpenhorn/db.py
+++ b/alpenhorn/db.py
@@ -118,11 +118,11 @@ def _connect(url=None, db=None):
 
 
 class EnumField(pw.Field):
-    """Implements an ENUM field for the peewee.
+    """Implements an ENUM field for peewee.
 
     Only MySQL and PostgreSQL support `ENUM` types natively in the database. For
-    Sqlite, the `ENUM` is implemented as an appropriately sized `VARCHAR` and
-    the validation is done at the Python level.
+    Sqlite (and others), the `ENUM` is implemented as an appropriately sized
+    `VARCHAR` and the validation is done at the Python level.
 
     .. warning::
         For the *native* ``ENUM`` to work you *must* register it with peewee like::
--- a/alpenhorn/db.py
+++ b/alpenhorn/db.py
@@ -104,11 +104,11 @@ def _connect(url=None, db=None):
     if url is not None:
         db = db_url.connect(url)
 
-    if isinstance(db, pw.SqliteDatabase):
-        EnumField = EnumFieldPython
-    else:
+    if isinstance(db, (pw.MySQLDatabase, pw.PostgresqlDatabase)):
         db.register_fields({'enum': 'enum'})
-        EnumField = EnumFieldDB
+        EnumField.native = True
+    else:
+        EnumField.native = False
 
     database_proxy.initialize(db)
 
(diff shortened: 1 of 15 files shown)


jrs65:
To get `alpenhornd` running, you can do the following. Creating a minimal configuration file `alpenhorn.conf`:

```yaml

database:
    url: sqlite:///testdb.db

extensions:
    - alpenhorn.generic

# Set any configuration for acquisition type extensions
acq_types:
    generic:
        patterns:
            - "**"

# Set any configuration for file type extensions
file_types:
    generic:
        patterns:
            - "*.h5"
            - "*.log"
```

Install alpenhorn in the usual way (e.g. `python setup.py develop`), then run the following in the shell:

```bash
$ export ALPENHORN_CONFIG_FILE=alpenhorn.conf
$ export ALPENHORN_LOG_FILE=/dev/null
$ alpenhorn init
$ alpenhornd
```

Note that it doesn't do anything at the moment. Neither the import, nor update code is implemented.

cubranic:
Looks great. I have a few questions I put in the review, hopefully that doesn't go away when I merge this, but don't want to hold it up any longer.

jrs65:
Hmmm. I'm not sure your comments ended up anywhere I can see them, which seems odd. Was it this PR you added them to? Just inline comments on the commits?

cubranic:
Yes, that’s where I could see them. I didn’t know I need to “submit review”. Now I have and you can see it in the PR conversation.

alpenhorn/db.py
```diff
@@ -15,93 +15,177 @@
...
+    ----------
+    url : str, optional
+        Database url using the scheme from `playhouse.db_url`.
+    db : `peewee.Database`, optional
+        Peewee database instance to connect alpenhorn to.
+    """
 
-class PropertyType(Exception):
-    """Bad property type."""
+    global EnumField
```
jrs65: Yeah. I guess we could remove than one.

alpenhorn/extensions.py
```diff
@@ -0,0 +1,146 @@
...
+    db : `peewee.Database`
+        A connected `peewee.Database` instance or `None` if there was no
+        database extension specified.
+    """
+
+    dbconnect = None
+
+    for ext_dict in _ext:
+
+        if 'database' in ext_dict:
```
jrs65: I figured we just needed one well specified order to do this, and last wins seemed easier and more intuitive. I agree that we probably should warn when this happens though.

tests/test_extensions.py
```diff
@@ -0,0 +1,91 @@
...
+        extensions.register_type_extensions()
+
+    # Check that we have registered every known type
+    acquisition.AcqType.check_registration()
+    acquisition.FileType.check_registration()
+
+    # Check that the correct entries have appeared
+    assert 'generic' in acquisition.AcqType._registered_acq_types
+    assert 'generic' in acquisition.FileType._registered_file_types
+    assert generic.GenericAcqInfo is acquisition.AcqType._registered_acq_types['generic']
```
jrs65: Is the question how come `_registered_acq_types` is not a list? Or how come `_registered_acq_types['generic']` is not a list? If the former, I think it probably could be. Each of the keys should be equal to the `_acq_type` of the the AcqInfo class. I mostly stored it as a dict to ease look up of by name, figuring that as there is meant to be no direct access to the dict I didn't need to worry about the key and the `_acq_type` somehow getting out of sync.

It is a bit inelegant though, so I'd be fine if you wanted to change it.

PR #16, opened 2017-05-03
Functional tests using Docker

jrs65:
This pull requests adds in the ability to test the behaviour of alpenhorn running across multiple systems. It uses docker to setup a test system of several instances of alpenhornd running against a separate MySQL system.

These tests are pretty lightweight for what they are, but compared to the unit tests they will take a lot more time.

To run the tests, simply install Docker (this was easy enough on my laptop using Docker for Mac), install the docker python SDK `pip install docker`, and set the environment variable `RUN_DOCKER_TESTS`.


jrs65:
I just changed how to enable/disable these tests to be via setting an environment variable. Hopefully this means travis won't break.

PR #17, opened 2017-05-05
Import and update loop

jrs65:
This major set of changes brings back the ability of alpenhorn to actually manage an archive!

Major changes:

 - Added a minimally modified version of the update loop code
   (excluding HPSS support).
 - Added back in the setup code for auto_import. With a few changes:
   * The code now derives its cache of previously imported files from
     the database.
   * Only the filesystem pollng watchdog is used, as the code to
     determine if the filesystem is local or network mounted is
     sketchy.
 - The client now starts a connection to the database and so is
   functional again (in theory).

More minor changes and cleanups:

 - Improved the configuration code with a better scheme for merging
   config files.
 - Fixed up the client tests to mock out the db connection/config
   loading code.
 - Changed verify command such that it gives a non zero exit status
   with a bad archive.
 - Changed the name of config dictionary.
 - Created a util module and moved various functions there.
 - Added tests for sync.
 - Fixed up the functional test for file import.

### DIFF
--- a/alpenhorn/auto_import.py
+++ b/alpenhorn/auto_import.py
@@ -4,16 +4,18 @@ from __future__ import division
 from __future__ import absolute_import
 import time
 import os
-import re
 
 import peewee as pw
 
+from watchdog.observers import Observer
+from watchdog.observers.polling import PollingObserver
 from watchdog.events import FileSystemEventHandler
 
-import alpenhorn.db as db
-import alpenhorn.archive as ar
-import alpenhorn.storage as st
-import alpenhorn.acquisition as ac
+from . import db
+from . import archive as ar
+from . import acquisition as ac
+from . import util
+from . import config
 
 # Setup the logging
 from . import logger
@@ -66,13 +68,6 @@ def _import_file(node, root, file_path):
         log.info("Skipping non-acquisition path %s." % file_path)
         return
 
-    # TODO: imported files caching
-    # if import_done is not None:
-    #     i = bisect.bisect_left(import_done, fullpath)
-    #     if i != len(import_done) and import_done[i] == fullpath:
-    #         log.debug("Skipping already-registered file %s." % fullpath)
-    #         return
-
     # Figure out which acquisition this is; add if necessary.
     acq_type, acq_name = acq_type_name
     try:
@@ -98,7 +93,7 @@ def _import_file(node, root, file_path):
 
     except pw.DoesNotExist:
         log.debug("Computing md5sum of \"%s\"." % file_name)
-        md5sum = md5sum_file(fullpath, cmd_line=False)
+        md5sum = util.md5sum_file(fullpath, cmd_line=False)
         size_b = os.path.getsize(fullpath)
 
         done = False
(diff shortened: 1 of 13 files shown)


alpenhorn/auto_import.py
```diff
@@ -310,3 +261,73 @@ def on_deleted(self, event):
...
+            q = (ar.ArchiveFileCopy.select(ac.ArchiveFile.name, ac.ArchiveAcq.name)
+                 .where(ar.ArchiveFileCopy.node == node,
+                        ar.ArchiveFileCopy.has_file == 'Y')
+                 .join(ac.ArchiveFile).join(ac.ArchiveAcq))
+
+            already_imported_files = [os.path.join(a, f) for a, f in q.tuples()]
+
+            log.info("Crawling base directory \"%s\" for new files.", node.root)
+
+            for acq_name, d, f_list in os.walk(node.root):
```
cubranic: Is this going to work with deeply-nested directory structure?

tests/docker/test_docker.py
```diff
@@ -291,28 +297,103 @@ def test_import(workers, test_files):
...
+    # auto_import, and then verify that they all got imported to the db
+    # correctly.
 
-            assert copy_query.count() == 1
-            copy_obj = copy_query.get()
+    # Create the files
+    _make_files(test_files, workers[0]['dir'], skip_lock=True)
+
+    # Wait for the auto_import to catch them (it polls at 30s intervals)
+    time.sleep(3)
```
cubranic: Is sleeping for 3 seconds enough when auto_import polls every 30s?

jrs65: The `alpenhorn.conf` used by the docker tests reduces the polling period down to 2s, so I think it's fine.

alpenhorn/update.py
```diff
@@ -0,0 +1,466 @@
...
+                                  .get()
+                        fcopy.has_file = 'Y'
+                        fcopy.wants_file = 'Y'
+                        fcopy.save()
+                        done = True
+                    except pw.OperationalError:
+                        log.error("MySQL connexion dropped. Will attempt to reconnect in "
+                                  "five seconds.")
+                        time.sleep(5)
+                        db.config_connect()
```
jrs65: I agree that this isn't great. It's not nearly as necessary since we started using the peewee RetryOperation mixin, or what ever it's called. Now that we've customised that we could implement slightly more complex logic too, with delays and more attempts to restart. I'm going to leave this for the moment, but add changing it to our todo list.

alpenhorn/update.py
```diff
@@ -0,0 +1,466 @@
...
+        fcopy.save()
+
+
+def update_node_delete(node):
+    """Process this node for files to delete."""
+
+    # If we are below the minimum available size, we should consider all files
+    # not explicitly wanted (wants_file != 'Y') as candidates for deletion,
+    # otherwise only those explicitly marked (wants_file == 'N')
+    # Also do not clean on archive type nodes.
```
jrs65: Yeah. That's unclear. We allow cleaning on archive nodes when there are enough copies elsewhere and they explicitly marked as `wants_file='N'`. I'll clarify the comment.

alpenhorn/auto_import.py
jrs65: Good question. I thought this was tested by the docker tests, but let me check that one.

alpenhorn/update.py
```diff
@@ -0,0 +1,466 @@
...
+
+            # If we couldn't just link the file, try copying it with rsync.
+            except OSError:
+                if util.command_available('rsync'):
+                    cmd = "rsync -%s %s %s" % (RSYNC_FLAG, from_path, to_dir)
+                    ret, stdout, stderr = util.run_command(cmd)
+
+                    md5sum = util.md5sum_file(to_file) if ret == 0 else None
+                else:
+                    log.warn("No commands available to complete this transfer.")
```
jrs65: Sure. I guess there's a few advantages to using rsync, mostly that it'll resume a failed transfer which could be useful. Let me look at changing it, as just using an internal stdlib would be nicer. Again, I'll add it as a todo list

alpenhorn/auto_import.py
jrs65: Hmmm. You're right that this shouldn't actually work at all. The enclosed call to `import_file` isn't actually correct. I think after that is fixed the `os.walk` part should be fine though, as the logic in `import_file` is meant to figure out the acq, file split.

jrs65:
I think this is about ready to merge, so I'm going to pull the trigger. I've moved the few suggested changes to be todo items.

PR #18, opened 2017-05-05
Tidied up the logging code

jrs65:
This pull request is a cleanup of the logging code, switching to getting a logger object at the module level direct from `logging`, and having `alpenhorn.logger` be only for configuring the logging output.

Most recommendations for python are to simply allow logging output to go to stdout and then let the system process it with its own tools (systemd, syslog etc). This is mostly fine, and is what should do by default, but on machines where alpenhorn doesn't run as a proper service (e.g. SciNet for CHIME), then we might need to specifically configure file locations. That ability has been removed in this PR, and so should be brought back later.

### DIFF
--- a/alpenhorn/service.py
+++ b/alpenhorn/service.py
@@ -5,13 +5,15 @@ from __future__ import absolute_import
 
 import sys
 import socket
+import logging
 
 import click
 
-from alpenhorn import (logger, extensions, db, config,
-                       auto_import, storage, update)
+from . import (extensions, db, config, logger,
+               auto_import, storage, update)
 
-log = logger.get_log()
+
+log = logging.getLogger(__name__)
 
 
 # Register Hook to Log Exception
@@ -31,6 +33,9 @@ def cli():
 
     # Load the configuration for alpenhorn
     config.load_config()
+
+    # Set up logging
+    logger.start_logging()
 
     # Attempt to load any alpenhor extensions
     extensions.load_extensions()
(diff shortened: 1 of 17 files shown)


jrs65:
I have no idea why the travis build broke after this change. It only removed a single commented line.

cubranic:
re: failing CI -- I was going to say that it could be a random problem with monkeypatching, but it happened on all three versions of Python. The one difference I see is that the failing one uses pytest 3.1.1, while the previous commit succeeded with pytest 3.0.7. (Also, `py` is 1.4.34 and 1.4.33, respectively. Should be a less significant change, though.)

cubranic:
Yep, I was able to reproduce the failure on a different branch after upgrading pyfakefs to 3.2

cubranic:
Moving `import os` out of `config/load_config` and at the module level fixes the problem. I think having the import done inside the function restored the 'os.path.exists' that's monkeypatched by `pyfakefs`. I'm not sure why it worked before.

alpenhorn/service.py
```diff
@@ -32,6 +34,9 @@ def cli():
     # Load the configuration for alpenhorn
     config.load_config()
 
+    # Set up logging
+    logger.start_logging()
+
```
jrs65: The logging isn't really finalised for the client at the moment. We need to decide how much logging should be displayed for the client, and how it should be configured.

jrs65:
Ah sweet. Thanks Davor! Glad you figured that one out.

PR #24, opened 2017-10-18
Bug with file selection in alpenhorn clean

tristpinsm:
Files marked as wants=M would not be selected by the clean command with the --now flag.
Please don't close my branch.


cubranic:
Can you also add those tests to "test_client.py"? Those can be run without docker.

tristpinsm:
I think everything is wrapped up now!

jrs65:
Looks great. Thanks for all your work!

PR #26, opened 2018-04-08
Node validation mechanism and import fixes

jrs65:
This pull request introduces a requirement that a node contains an `ALPENHORN_NODE` file at its root directory, otherwise it will be unmounted.

These changes exposed some issues with the `auto_import` functionality, so these have been fixed. Notably, `import_file` now only takes two arguments the `node` object and a path to the file (the `root_path` argument was redundant and has been removed).

### DIFF
--- a/alpenhorn/auto_import.py
+++ b/alpenhorn/auto_import.py
@@ -181,21 +181,11 @@ def add_acq(acq_type, name, node, comment=""):
         raise AlreadyExists("Acquisition \"%s\" already exists in DB." %
                             name)
 
-    # At the moment we need an instrument, so just create one.
-    # TODO: this should probably go away
-    try:
-        inst_rec = ac.ArchiveInst.get(name='inst')
-    except pw.DoesNotExist:
-        print("Creating inst.")
-        inst_rec = ac.ArchiveInst.create(name='inst')
-
     # Create the ArchiveAcq entry and the AcqInfo entry for the acquisition. Run
     # in a transaction so we don't end up with inconsistency.
     with db.database_proxy.atomic():
         # Insert the archive record
-        acq = ac.ArchiveAcq.create(name=name, inst=inst_rec,
-                                   type=acq_type,
-                                   comment=comment)
+        acq = ac.ArchiveAcq.create(name=name, type=acq_type, comment=comment)
 
         # Generate the metadata table
         acq_type.acq_info.new(acq, node)
--- a/alpenhorn/auto_import.py
+++ b/alpenhorn/auto_import.py
@@ -66,8 +66,6 @@ def _import_file(node, file_path):
         be within the node root, otherwise is is assumed to be relative to
         the node root.
     """
-    # global import_done
-    curr_done = True
 
     log.debug("Considering \"%s\" for import.", file_path)
 
--- a/alpenhorn/auto_import.py
+++ b/alpenhorn/auto_import.py
@@ -287,8 +287,8 @@ def setup_observers(node_list):
                     if file_name in already_imported_files:
                         log.debug("Skipping already-registered file %s.", file_name)
                     else:
-                        import_file(node, node.root, os.path.basename(acq_name),
-                                    file_name)
+                        import_file(node, node.root, os.path.join(acq_name,
+                                    file_name))
 
             # TODO: this test needs fixing
             # If it is an NFS mount, then the default Observer() doesn't work.
--- a/alpenhorn/auto_import.py
+++ b/alpenhorn/auto_import.py
@@ -9,7 +9,6 @@ import logging
 
 import peewee as pw
 
-from watchdog.observers import Observer
 from watchdog.observers.polling import PollingObserver
 from watchdog.events import FileSystemEventHandler
 
@@ -23,11 +22,11 @@ from . import config
 log = logging.getLogger(__name__)
 
 
-def import_file(node, root, file_path):
+def import_file(node, file_path):
     done = False
     while not done:
         try:
-            _import_file(node, root, file_path)
+            _import_file(node, file_path)
             done = True
         except pw.OperationalError as e:
             log.exception(e)
@@ -38,7 +37,16 @@ def import_file(node, root, file_path):
             db.database_proxy.connect()
 
 
-def _import_file(node, root, file_path):
+def in_directory(file, directory):
+    """Test if file is contained within the directory. Does not check existence."""
+    directory = os.path.join(directory, '')
+
+    #return true, if the common prefix of both is equal to directory
+    #e.g. /a/b/c/d.rst and directory is /a/b, the common prefix is /a/b
+    return os.path.commonprefix([file, directory]) == directory
+
+
+def _import_file(node, file_path):
     """Import a file into the DB.
 
     This routine adds the following to the database, if they do not already exist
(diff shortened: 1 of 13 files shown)


alpenhorn/auto_import.py
```diff
@@ -48,52 +56,81 @@ def _import_file(node, root, file_path):
...
+    if os.path.isfile(os.path.join(dir_name, ".%s.lock" % base_name)):
+        log.debug("Skipping \"%s\", which is locked by ch_master.py.", file_path)
         return
 
     # Check if we can handle this acquisition, and skip if we can't
-    acq_type_name = ac.AcqType.detect(file_path, node)
+    acq_type_name = ac.AcqType.detect(relpath, node)
     if acq_type_name is None:
-        log.info("Skipping non-acquisition path %s." % file_path)
+        log.info("Skipping non-acquisition path \"%s\".", file_path)
```
cahofer: should this say relpath? At least that's the one you checking in AcqType.detect

jrs65: Hmmmm. That's a good point. I think it probably should stay as `file_path`, basically as it's good to give the original path (before it's been processed internally) so it's easier for the user to figure out what went wrong.

alpenhorn/auto_import.py
```diff
@@ -48,52 +56,79 @@ def _import_file(node, root, file_path):
...
+    # Ensure the path is an absolute path within the node
+    if os.path.isabs(file_path):
+        if not in_directory(file_path, node.root):
+            log.error("File \"%s\" was not an absolute path within the node \"%s\"", file_path, node.root)
+            return
+    else:
+        file_path = os.path.join(node.root, file_path)
+    abspath = os.path.normpath(file_path)
+
+    # Skip requests to import a directory. Again these are occasionally sent by the watchdog
```
cubranic: This could probably be the very first check in the function, to save us the effort of checking for inclusion in the root, normalizing, etc., when there is nothing that can be done with it.

alpenhorn/update.py
```diff
@@ -64,12 +64,18 @@ def update_node(node):
...
-        log.debug("Skipping suspected node \"%s\"." % node.name)
+        log.debug("Skipping suspected node \"%s\".", node.name)
 
-    log.info("Updating node \"%s\"." % (node.name))
+    log.info("Updating node \"%s\".", node.name)
+
+    # Check if the node is acutally mounted in the filesystem
+    check_node = update_node_mounted(node)
+
+    if check_node is False:
```
cubranic: Isn't `if not check_node` considered more Pythonic? (And would continue to work if we changed update_node_mounted to return `None` instead of `False` on failure.)

Also, I'm waffling whether to recommend that `check_node` be renamed to "`is_mounted`" or just be replaced entirely with:
```py
if not update_node_mounted(node):
    return
```

tests/docker/test_docker.py
```diff
@@ -464,6 +472,30 @@ def test_clean(workers, network, test_files):
...
+
+    os.rename(data_dir0 + '/DIFFERENT_NAME', data_dir0 + '/ALPENHORN_NODE')
+
+    time.sleep(3)
+
+    node_0 = st.StorageNode.get(name=this_node.name)
+    node_0.mounted = True
+    node_0.save(only=node_0.dirty_fields)
+
+    assert node_0.mounted
```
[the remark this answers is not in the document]
cubranic: Instead, we could update it in the database, wait for 3s, and check that it's still mounted, like this:
```py
    os.rename(data_dir0 + '/DIFFERENT_NAME', data_dir0 + '/ALPENHORN_NODE')

    node_0 = st.StorageNode.get(name=this_node.name)
    node_0.mounted = True
    node_0.save(only=node_0.dirty_fields)

    time.sleep(3)

    node_0 = st.StorageNode.get(name=this_node.name)
    assert node_0.mounted
```

cubranic:
Yes, I believe so. An admin user can always mount and unmount by interacting with the OS directly, e.g., with the shell commands, in which case Alpenhorn’s database would be out of sync with the filesystem. This goes for both mounting and unmounting.

alpenhorn/update.py
```diff
@@ -89,6 +95,44 @@ def update_node(node):
...
+        else:
+            log.error("Node \"%s\" is not mounted as expected from db (missing %s file).",
+                      node.name, fname)
+
+    # If we are here the node it not correctly mounted so we should unmount it...
+    node.mounted = False
+    node.save(only=node.dirty_fields)  # save only fields that have been updated
+    log.info("Correcting. Node %s is now set to unmounted", node.name)
+
+    return node.mounted
```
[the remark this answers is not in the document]
cubranic: Yes, I believe so. An admin user can always mount and unmount by interacting with the OS directly, e.g., with the shell commands, in which case Alpenhorn’s database would be out of sync with the filesystem. This goes for both mounting and unmounting.

cubranic:
Also, the check will fail if ALPENHORN_NODE has a newline character at the end of the first line. (And it will, if you create it with `echo node1 > ALPENHORN_NODE` or with `vi`.)

cubranic:
It looks like the check for ALPENHORN_NODE doesn't happen until the update loop starts. However, when alpenhornd starts up, it will first call auto_import.setup_observers, which will automatically import all new files in the node root, regardless of whether ALPENHORN_NODE is present or not.

```
Attaching to bcnetdemo_host_1_1
host_1_1  | Apr 18 23:05:30 INFO >> Loading config file /etc/alpenhorn/alpenhorn.conf
host_1_1  | Apr 18 23:05:30 INFO >> Loading extension custom_test_types
host_1_1  | Apr 18 23:05:30 INFO >> Registering new acquisition type zab
host_1_1  | Apr 18 23:05:30 INFO >> Registering new acquisition type quux
host_1_1  | Apr 18 23:05:30 INFO >> Registering new file type zxc
host_1_1  | Apr 18 23:05:30 INFO >> Registering new file type log
host_1_1  | Apr 18 23:05:30 INFO >> Crawling base directory "/data" for new files.
host_1_1  | Apr 18 23:05:30 INFO >> Crawling "/data".
host_1_1  | Apr 18 23:05:30 INFO >> Crawling "/data/12345678T000000Z_inst_zab".
host_1_1  | Apr 18 23:05:30 INFO >> Acquisition "12345678T000000Z_inst_zab" added to DB.
host_1_1  | Apr 18 23:05:30 INFO >> File "12345678T000000Z_inst_zab/ch_master.log" added to DB.
host_1_1  | Apr 18 23:05:30 INFO >> Registered file copy "12345678T000000Z_inst_zab/ch_master.log" to DB.
host_1_1  | Apr 18 23:05:30 INFO >> Registering node "node_1" for auto_import watchdog.
host_1_1  | Apr 18 23:05:30 INFO >> Updating node "node_1".
host_1_1  | Apr 18 23:05:30 ERROR >> Node "node_1" is not mounted as expected from db (missing ALPENHORN_NODE file).
host_1_1  | Apr 18 23:05:30 INFO >> Correcting. Node node_1 is now set to unmounted
```

alpenhorn/auto_import.py
cubranic: @jrs65 thoughts?

alpenhorn/update.py
cubranic: @jrs65 what do you think? Keep "`check_node`", rename to "`is_mounted`", or eliminate and use "`if not update_node_mounted(node)`" directly, since the variable is not used anywhere else in the method.

cubranic:
I rebased this branch on master, which has had quite a lot of small changes in the last week. Any last comments / responses to my comments, before I merge?

jrs65:
Oh yeah. Sorry. I was going to do the fixes that you suggested. Let me do it now then they don't get lost.

alpenhorn/auto_import.py
jrs65: Yeah, this one makes sense to change around. I'll change this one.

jrs65:
I'll add a fix for the new line check issue.

jrs65:
I'm going to ignore the auto_import issue for now. I think we should refactor the starting/stopping of the auto_import observers into the main loop as that will solve the other issues you were talking about.

alpenhorn/auto_import.py
jrs65: Actually. I remember this now. The reason to not swap them around is that you might need to resolve the path relative to the node root before you can test that it's a directory or not.

cubranic:
Ouch, that merge of my rebase made it hurt to look at the history. My eyes thank you for the "squash and merge", GitHub!

Issue #28, opened 2018-04-08
Ensure that it works with the newest docker python api


jrs65:

cubranic:
From the change log for v3: 
> `Container.exec_run` now returns a tuple of (exit_code, output) instead of just the output.

The current release is 3.2.1.

PR #30, opened 2018-04-18
#28: use docker-py v3 API

cubranic:
Since v3, `Container.exec_run` returns a tuple of (exit_code, output)`.

We could also check `docker.__version__` after importing, and raise an exception if it's not `=~ 3.0`.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,3 +1,4 @@
 pytest >= 2.10
 pyfakefs >= 3.1
 mock; python_version < '3.0'
+docker >= 3.0
--- a/tests/docker/test_docker.py
+++ b/tests/docker/test_docker.py
@@ -342,7 +342,7 @@ def _verify_files(worker):
     )
 
     # Convert the output back to an exit status
-    assert not int(output)
+    assert not output.exit_code
 
 
 # ====== Test the auto_import behaviour ======


jrs65:
Do we also need to require a modern enough version of docker-py for this to work?

cubranic:
That’s what I meant by “docker”, that’s the name of the imported package.

jrs65:
Sorry. I was just thinking that maybe we should update the `test-requirements.txt` to require a recent enough version. However, I notice now that docker isn't actually in there? Should we put it in? It's not like it forces you to actually install the full blown docker.

cubranic:
It looks like the check for ALPENHORN_NODE doesn't happen until the update loop starts. However, when alpenhornd starts up, it will first call `auto_import.setup_observers`, which will automatically import all new files in the node root, regardless of whether ALPENHORN_NODE is present or not.

```
Attaching to bcnetdemo_host_1_1
host_1_1  | Apr 18 23:05:30 INFO >> Loading config file /etc/alpenhorn/alpenhorn.conf
host_1_1  | Apr 18 23:05:30 INFO >> Loading extension custom_test_types
host_1_1  | Apr 18 23:05:30 INFO >> Registering new acquisition type zab
host_1_1  | Apr 18 23:05:30 INFO >> Registering new acquisition type quux
host_1_1  | Apr 18 23:05:30 INFO >> Registering new file type zxc
host_1_1  | Apr 18 23:05:30 INFO >> Registering new file type log
host_1_1  | Apr 18 23:05:30 INFO >> Crawling base directory "/data" for new files.
host_1_1  | Apr 18 23:05:30 INFO >> Crawling "/data".
host_1_1  | Apr 18 23:05:30 INFO >> Crawling "/data/12345678T000000Z_inst_zab".
host_1_1  | Apr 18 23:05:30 INFO >> Acquisition "12345678T000000Z_inst_zab" added to DB.
host_1_1  | Apr 18 23:05:30 INFO >> File "12345678T000000Z_inst_zab/ch_master.log" added to DB.
host_1_1  | Apr 18 23:05:30 INFO >> Registered file copy "12345678T000000Z_inst_zab/ch_master.log" to DB.
host_1_1  | Apr 18 23:05:30 INFO >> Registering node "node_1" for auto_import watchdog.
host_1_1  | Apr 18 23:05:30 INFO >> Updating node "node_1".
host_1_1  | Apr 18 23:05:30 ERROR >> Node "node_1" is not mounted as expected from db (missing ALPENHORN_NODE file).
host_1_1  | Apr 18 23:05:30 INFO >> Correcting. Node node_1 is now set to unmounted
```

cubranic:
Agreed!

PR #66, opened 2019-01-18
Refactor test fixtures to work in Pytest v4

cubranic:
Pytest deprecated direct calling of fixtures in release 4. Instead, the
[recommended pattern][1] is to refactor the "callable" part of the fixture into a
regular function, and call that from any fixture that needs it.

[1]: https://docs.pytest.org/en/latest/deprecations.html#calling-fixtures-directly

### DIFF
--- a/tests/test_acquisition_model.py
+++ b/tests/test_acquisition_model.py
@@ -19,12 +19,10 @@ from alpenhorn.acquisition import *
 
 tests_path = path.abspath(path.dirname(__file__))
 
-@pytest.fixture
-def fixtures(clear_db=True):
-    """Initializes an in-memory Sqlite database with data in tests/fixtures"""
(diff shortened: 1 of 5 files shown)


cubranic:
This needs to be rebased on #67 to pass, so let's merge that one first.

cubranic:
Seems like Github's CI uses a pre-v4 Pytest, as this wasn't being flagged. But I just did a fresh conda install on my laptop, and the current version of Pytest 4.1.1.

jrs65:
@cubranic do you want to bump the version of pytest in the `test-requirements.txt` file to `>= 4.0.0` or higher.

cubranic:
@jrs65 It is backward compatible, so I don't think we need to require it.

PR #67, opened 2019-01-19
Always pass a status code > 0 when performing an error exit

cubranic:


cubranic:
@jrs65 @ketiltrout Sounds reasonable?

PR #72, opened 2019-01-25
Add missing method argument `node_root` to AcqInfoBase.set_info

cubranic:
The calling code invokes `AcqInfo.set_info()` with both `acqpath` and
`node_root`, but the abstract method in AcqInfoBase is missing the 'node_root'.


jrs65:
Good catch.

PR #73, opened 2019-01-25
Use short hostname consistently as the default (fixes #71)

### LINKED ISSUE #71: Default hostname differs in `create-node` and `mounted`
My laptop's hostname is `davit.phas.ubc.ca`, as reported by `/bin/hostname`. That's the hostname that's used by `alpenhorn create-node` when I don't specify the `--hostname` option. However, `alpenhornd` and `alpenhorn mounted` see the hostname as being just `davit`. It's a minor problem, but confusing to a new user, and annoying when just trying to quickly create a node on the same host where the daemon is already running.

cubranic:


jrs65:
Great. This is a good one to have fixed up.

PR #74, opened 2019-01-29
Increase the limit for file name to 255 characters

cubranic:
Since we allow nested directories inside acquisitions, file names can easily get
longer than the current 64-character limit. Raising this to 255 is at least in
line with filename limit on common Linux filesystems, although even that is not
quite enough for any possibility, as the maximum *path* length in current Linux
is usually 4,096 bytes.


jrs65:
Sounds good. We might need @ketiltrout to add this to his migration plan.

PR #77, opened 2019-01-31
Restore RetryOperationalError mixing (fix #76)

### LINKED ISSUE #76: Check that db.RobustProxy works as intended
While checking that RobustProxy still works in Peewee 3, I couldn't get it to actually trigger. As I investigated the source code, my guess is that it has to be used as a mixin for the Database object ([doc](http://docs.peewee-orm.com/en/2.10.2/peewee/database.html#automatic-reconnect)), not a subclass of `peewee.Proxy`.

The reason for this is that when we invoke an operation, e.g., `StorageNode.create_table`, it ends up in the `execute` method of the proxied Database connection. This calls `self.execute_sql()` (src [peewee.py#L2734](https://github.com/coleifer/peewee/blob/master/peewee.py#L2734)), where `self` is the proxied Database instance, not the RobustProxy proxier.

If this is indeed the case (@jrs65 can you check?), then the solution is probably to convert RobustProxy into a [mixin](https://github.com/coleifer/peewee/issues/1472#issuecomment-364327161) like the old Peewee `RetryOperationalError`, and dynamically add that mixin type to the result of `db_url.connect()` if it's a `Mysql` or `PostgresqlDatabase`. ([This is how](https://stackoverflow.com/a/8545134).) Then `db.database_proxy` would go back to being an instance of`peewee.Proxy`.

cubranic:
RobustProxy doesn't work as intended, because to really override `Database.execute_sql` it has to be used as a mixin for the Database object, not a subclass of peewee.Proxy.

This commit restores the old `RetryOperationalError`, updated for the current Peewee APIs, and dynamically adds its type to the result of db_url.connect(). `db.database_proxy` goes back to being an instance of `peewee.Proxy`.

### DIFF
--- a/alpenhorn/db.py
+++ b/alpenhorn/db.py
@@ -15,40 +15,7 @@ logger = logging.getLogger(__name__)
 
 # _property = property  # Do this because we want a class named "property".
 
-class RobustProxy(pw.Proxy):
-    """A robust database proxy.
-
-    A merger of peewee's database proxy support and the functionality of the
-    `RetryOperationalError` mixin. to try and make the database connection more
-    robust by default.
-    """
-
-    def execute_sql(self, sql, params=None, require_commit=True):
-        # Code in this function is a modified version of some in peewee
-        # Copyright (c) 2010 Charles Leifer
-
-        # Check that the proxy exists
-        if self.obj is None:
-            raise AttributeError('Cannot use uninitialized Proxy.')
-
-        # If it does try to run the query...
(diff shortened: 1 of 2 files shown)


cubranic:
(merge after #78)

cubranic:
Does this look reasonable to you @jrs65? I tried to exercise the feature by mocking a failing connection in test_db.py unit tests, and it seems to work as intended.

PR #84, opened 2019-02-28
Fix #59: Add DB constraint for only one copy of a file per node

### LINKED ISSUE #59: Missing constraint on ArchiveFileCopy
There should be a unique key constraint added to `ArchiveFileCopy` along the lines of 
```
pw.CompositeKey('file','node')
```
to enforce only one copy of a file per node.

cubranic:


jrs65:
Thanks @cubranic. Very much appreciate all this work you're doing on alpenhorn. It's going to be solid as a rock.

cubranic:
Do you understand why it works with a unique index but not a composite primary key?

PR #91, opened 2019-03-03
Add a check for ALPENHORN_NODE file when mounting the node (fix #7)

### LINKED ISSUE #7: Add a check that a node is actually mounted on filesystem
An issue we occasionally have with CHIME is that is alpenhorn gets start and a node which is marked as `mounted` in the database is not actually mounted in the filesystem. When this happens any transfers requested from that node end up failing and causing files to be marked as missing.

One idea might be to have a marked in each node the alpenhorn can check for in the directory. Something simple as a requirement to have an ALPENHORN_NODE file sit in the root of the node.

Thoughts? @cubranic

cubranic:
This was already used in the Alpenhorn service to detect the case when a remote
mount went away. We now do the same check in the client command, so that it
prevents mistakenly mounting a storage node under another node's mountpoint.

### DIFF
--- a/alpenhorn/client.py
+++ b/alpenhorn/client.py
@@ -757,6 +757,13 @@ def mount(name, path, user, address, hostname):
         click.echo("Node \"%s\" is already mounted." % name)
         return
 
+    if path is not None:
+        node.root = path
+
+    if not util.alpenhorn_node_check(node):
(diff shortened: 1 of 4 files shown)


jrs65:
Awesome. This is definitely a good check to have.

PR #98, opened 2019-03-11
Create unit tests for steps in the update loop

cubranic:
This would have caught the runtime error fixed in the previous commit, without
having to run the full Docker tests, which don't get run during CI.


jrs65:
These are great!

cubranic:
I mostly adapted them over from the Docker ones, thanks for those!

PR #102, opened 2019-03-18
Add an option to `create_node` for setting the username (fix #39)

### LINKED ISSUE #39: `create_node` should include the option for setting StorageNode.username
In the docker test, we create nodes with:

```py
st.StorageNode.create(
  name=('node_%i' % i), root='/data', username='root',
  group=group, host=hostname, address=hostname, mounted=True,
  auto_import=(i == 0), min_avail_gb=0.0
)
```

For consistency, we should be able to set do the same from the command line with `alpenhorn create_node`.

(To be fair, setting the username is still doable from the command line with `alpenhorn mount`.)

cubranic:


cubranic:
Given what @ketiltrout said, and since the code needs to be reworked since this functionality got moved into the `node` subcommand, I'll close this PR and open #129 to discuss the design.

PR #103, opened 2019-03-18
Improve user-friendliness of the `clean` command

cubranic:
- improves the help text and progress messages to make it clearer what's being cleaned means
- have a way to undo a "clean" request
- signal user errors with non-zero exit code

### DIFF
--- a/alpenhorn/client.py
+++ b/alpenhorn/client.py
@@ -438,14 +438,18 @@ def clean(node_name, days, cancel, force, now, target, acq):
     Files will never be removed until they are available on at least two
     archival nodes.
 
-    If --target is specified we will only remove files already available in the
-    TARGET_GROUP. This is useful for cleaning out intermediate locations such as
-    transport disks.
-
     Normally, files are marked to be removed only if the disk space on the node
     is running low. With the --now flag, they will be made available for
     immediate removal. Either way, they will *never* be actually removed until
     there are sufficient archival copies.
+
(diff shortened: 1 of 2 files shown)


cubranic:
@jrs65 If you agree with how the new "clean --cancel" feature operates, I will add the same mechanism to "sync".

jrs65:
Oh this is a great feature. Implementation looks good too.

alpenhorn/client.py
```diff
@@ -425,47 +425,72 @@ def verify(node_name, md5, fixdb, acq):
...
 @click.option('--now', '-n', help='Force immediate removal.', is_flag=True)
 @click.option('--target', metavar='TARGET_GROUP', default=None, type=str,
               help='Only clean files already available in this group.')
 @click.option('--acq', metavar='ACQ', default=None, type=str,
               help='Limit removal to acquisition.')
-def clean(node_name, days, force, now, target, acq):
+def clean(node_name, days, cancel, force, now, target, acq):
     """Clean up NODE by marking older files as potentially removable.
 
+    Files will never be removed until they are available on at least two
```
jrs65: In this documentation you might want to have a longer comment about the effects of the `--cancel` option. Given it completely inverts the operation that is to be done, we might want to discuss a little more.

cubranic: I was going to and then forgot to do it before committing. I'll add now.

Issue #63, opened 2018-11-22
Drop n_requests from ArchiveFileCopyRequest


ketiltrout:
Do we use this column now?  Will we at all if we implement #57 ?

ketiltrout:
Other than incrementing it, not used by alpenhorn, so should be fine to remove.

Issue #57, opened 2018-08-06
Change ArchiveFileCopyRequest to not be used to track multiple instances of the same request


cubranic:
Copied from a discussion in CHIME Infrastructure: 

> We would then have separate timestamps for: time requested, time transfer started, and time transfer ended.

jrs65:
Yup. Obviously we'd need to re-key the table.

ketiltrout:
Whose responsibility will it be to ensure there's only one pending request for a given `(file,node_from,group_to)`? The process adding new requests?

What is the intended behaviour for `alpenhornd`?  Is it going to check for pending duplicates or just assume there aren't any?

cubranic:
Bump. I forgot why we care about this. Is it just for the audit trail?

cubranic:
I talked about this with @jrs65 two days ago, but I still don't have the full spec clear in my head. Richard, can you clarify a few questions I still have:

1. when there is a pending copy request (i.e., not cancelled and not completed), and another `alpenhorn sync` is issued, what should happen?

2. if pending requests should be unique on `(file, node_from, group_to)`, can we have a database constraint to ensure it?

3. if we can't express this as a DB constraint, do we care, and do we try to make both Alpenhornd and the client code maintain the invariant?

jrs65:
> I talked about this with @jrs65 two days ago, but I still don't have the full spec clear in my head. Richard, can you clarify a few questions I still have:
> 
> 1. when there is a pending copy request (i.e., not cancelled and not completed), and another `alpenhorn sync` is issued, what should happen?

I would say that nothing should happen at this point. The client should probably report the number of requests that already existed.

> 2. if pending requests should be unique on `(file, node_from, group_to)`, can we have a database constraint to ensure it?

Yeah, that's an interesting one. That one is only unique when it's pending now. I don't know if there is a way of adding a conditional constraint like that.

> 3. if we can't express this as a DB constraint, do we care, and do we try to make both Alpenhornd and the client code maintain the invariant?

I think we should attempt to make them preserve that. That's probably a case of the client checking carefully for the existence of the request before inserting (probably expressed as an upsert), and the server marking *all* pending copies complete (or maybe the last one complete, and all others cancelled) when it finishes a transfer.

jrs65:
[...]

PR #104, opened 2019-03-22
Finer-grained ArchiveFileCopyRequest tracking and accounting

cubranic:
This implements schema changes proposed in #57 and #63, namely:

Instead having ArchiveFileCopyRequest have a single instance for each `(file, node_from, group_to)` combination, and updating them in place as a new copy operation is requested after being completed or cancelled once, this changetracks the copying at a request level by:

1. adding new columns for start and end time of the transfer

2. once a copy request is completed or cancelled, requesting the copying of a file again will create a new instance of a copy request for the same `(file, node_from, group_to)`.

With this semantics of copy requests, the `n_requests` column becomes redundant and is removed.

### DIFF
--- a/alpenhorn/archive.py
+++ b/alpenhorn/archive.py
@@ -60,7 +60,11 @@ class ArchiveFileCopyRequest(base_model):
     cancelled : bool
         Set to true if the copy is no longer wanted.
     timestamp : datetime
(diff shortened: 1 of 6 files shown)


jrs65:
Looks great. Thanks @cubranic

PR #105, opened 2019-03-25
Add a registered time to the ArchiveFile table (closes #56)

### LINKED ISSUE #56: Adding a registered time to the ArchiveFile table
Copied from a discussion in CHIME Infrastructure: 
> That would make some sense as not all files are timestream type files with a well defined start and end time.

cubranic:


alpenhorn/acquisition.py
```diff
@@ -297,12 +298,15 @@ class ArchiveFile(base_model):
...
         md5 checksum of file. Used for verifying integrity.
+    registered : datetime
+        The time the file was registered in the database.
     """
     acq = pw.ForeignKeyField(ArchiveAcq, backref='files')
     type = pw.ForeignKeyField(FileType, backref='files')
     name = pw.CharField(max_length=255)
     size_b = pw.BigIntegerField(null=True)
     md5sum = pw.CharField(null=True, max_length=32)
+    registered = pw.DateTimeField(default=datetime.datetime.now())
```
jrs65: I think this needs to be `datetime.datetime.now` (i.e. no `()`). If you do what's listed here it only evaluates the time at module import time (I think), however, if you pass a callable to peewee it will evaluate it when it constructs an instance.

jrs65: You might want to tighten up your unit test such that it would have caught this I guess. Maybe add a `sleep` to make sure that the module import time isn't within the 5 second window.

cubranic: You were right, good catch!

PR #107, opened 2019-03-25
Add true size field to ArchiveFileCopy (closes #58)

### LINKED ISSUE #58: Add a file size field to ArchiveFileCopy
This would record the "true" size, as used by the filesystem. (E.g., accounting for holes in sparse files or Lustre filesystem compression.) @ketiltrout is planning on adding it to alpenhorn1, and we should port it here as well.

cubranic:
This reflects the "true" size of the file, as used by the filesystem. (E.g.,
accounting for holes in sparse files or Lustre filesystem compression.)

### DIFF
--- a/alpenhorn/archive.py
+++ b/alpenhorn/archive.py
@@ -30,11 +30,15 @@ class ArchiveFileCopy(base_model):
         - 'M': maybe, can delete if we need space
         - 'N': no, attempt to delete
         In all cases we try to keep at least two copies of the file around.
+    size_b : integer
+        Allocated size of file in bytes (calculated as a multiple of 512-byte
+        blocks).
     """
     file = pw.ForeignKeyField(ArchiveFile, backref='copies')
     node = pw.ForeignKeyField(StorageNode, backref='copies')
     has_file = EnumField(['N', 'Y', 'M', 'X'], default='N')
     wants_file = EnumField(['Y', 'M', 'N'], default='Y')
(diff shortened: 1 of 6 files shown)


cubranic:
BTW Don, I update the true file size field not just when the copy is created, but also if its MD5 is rechecked after it was marked "has_file=M".

cubranic:
Rebased on current master. @ketiltrout can you have a look before I merge this?

alpenhorn/archive.py
```diff
@@ -30,11 +30,14 @@ class ArchiveFileCopy(base_model):
         - 'M': maybe, can delete if we need space
         - 'N': no, attempt to delete
         In all cases we try to keep at least two copies of the file around.
+    size_b : integer
+        Size of file in bytes.
```
ketiltrout: You should point out that this is the _actual_ size, not the _apparent_ size.  Maybe:
```
Actual size of file on disk in bytes.
```
or
```
Allocated size of file on disk in bytes.
```
Is it worth point out that this will be a multiple of 512?

cubranic: Agreed on both. This is calculated using `os.stat.st_blocks`, which the doc says is the "number of 512-byte blocks allocated for file", so our description could be 
```
allocated size of file on disk in bytes (calculated as a multiple of 512-byte blocks)
```
Sounds OK?

cubranic: Updated as above.

PR #110, opened 2019-06-06
Add an option for client 'import_files' to perform auto-import

cubranic:
This allows us to import files on a field node without having to run the daemon, or if we want to be selective about which directories to import.

### DIFF
--- a/alpenhorn/client.py
+++ b/alpenhorn/client.py
@@ -866,8 +866,9 @@ def deactivate(root_or_name):
 @click.argument('node_name', metavar='NODE')
 @click.option('-v', '--verbose', count=True)
 @click.option('--acq', help='Limit import to specified acquisition directories.', multiple=True, default=None)
+@click.option('--create', help='Allow import of new acquisitions', is_flag=True)
 @click.option('--dry', '-d', help='Dry run. Do not modify database.', is_flag=True)
-def import_files(node_name, verbose, acq, dry):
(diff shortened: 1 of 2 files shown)


cubranic:
OK, this is cleaned up, rebased on master, and ready for review. Note that only 35862104f29212c50dc92f0268e1ac7a9b7a7388 is a new commit, it looks like Github's UI got confused about which commits are part of this branch, probably after my rebases and force-pushes.

cubranic:
I agree and meant to ask about that as part of the review. I didn't use "register" because the comments seem to use "register" for adding ArchiveFileCopy, whereas this would register new acqs and ArchiveFile instances.

If that's not a problem, I do like "register-new". Should it also have a short name or as a somewhat unusual case require using the long one?

jrs65:
Register new seems good. Ideally a short name too, but the obvious candidates `-r` and `-n` typically have other meanings (i.e. recursive, and dry-run), so maybe don't bother.

cubranic:
It could be a capital letter, so "-N" would be safe.

Incidentally, the short name for `--dry` in its only use in our code is "-d", while as you say "-n" is the usual one. (We also use "-n" for "--now" in "clean" and "--nice" in "sync", but maybe it's not a big deal as they're distinct commands.)

cubranic:
I'm going to merge with just the "register-new", and then close this PR. We can review the short names across all the sub-commands for consistency and open a separate PR for whatever needs renaming or adding.

PR #116, opened 2019-06-14
Revise mechanics in `client.format_transport`

cubranic:
Using parted seems to trigger a reload of udev rules and so the device file goes missing briefly after it returns. This means that a) we should only use parted if we really have to; and b) include a short sleep after it to let udev to recreate the device files so the next process (e2label or mkfs) can find it.

Also, it's now recommended to use functions in module `subprocess` to run subprocesses instead of `os.system` and `os.popen`/`popen3`.

### DIFF
--- a/alpenhorn/client.py
+++ b/alpenhorn/client.py
@@ -668,8 +668,8 @@ def format_transport(serial_num):
 
     # Figure out if it is formatted.
     print("Checking to see if disc is formatted. Please wait.")
+    formatted = False
     try:
-        formatted = False
         # check if the block device is partitioned
         subprocess.check_output(['blkid', '-p', dev])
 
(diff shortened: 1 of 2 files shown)


alpenhorn/client.py
```diff
@@ -666,26 +668,40 @@ def format_transport(serial_num):
...
-    part_start = False
-    while True:
-        l = fp.readline()
-        if not l:
-            break
-        if l.find("Number") == 0 and l.find("Start") > 0 and l.find("File system") > 0:
-            part_start = True
-        elif l.strip() != "" and part_start:
+    try:
+        formatted = False
```
ketiltrout: Since we can assume this assignment will never raise `subprocess.CalledProcessError`, it should be done outside the try block.

PR #123, opened 2019-12-20
Break up CLI into sub-commands grouped by target type (closes #122)

### LINKED ISSUE #122: Break up client into `noun verb` subcommands
As discussed on Slack, the current list of client commands is getting long, and it would only get worse if we added more. (And we do, for instance for inspecting or modifying various minor aspects of the objects in the database -- e.g., listing the nodes' names, or changing their description, etc. -- without needing to do it in SQL.) The natural solution would be to break it down into subcommands, probably of the "noun verb" variety: `node list`, `node create`, `node modify`, etc. (This is similar to what we already do in data-flagging tools.)

There is the question of whether to use "noun verb" or "verb noun" pattern for naming. I don't know if there is a generally accepted principle: for example Git offers "noun verb", when it does have subcommands (e.g., "branch" or "remote"). I think in this case using "noun verb" works a better, just because Nodes, Acquisitions, and ArchiveFiles objects are so central, and everything else about the code revolves around them. With that in mind, here is a proposal:

- node: create, list, modify, status(?)
- acq(uisition): import, list, info, (?)
- archive/file: ...
- transport: list, mount, unmount?

Do we need to keep all the existing commands, for backward compatibility? With a plan to deprecate them?

cubranic:
Current organization:

- node:
  - new: list, rename, modify
  - rearranged: create, activate, deactivate, verify, clean

- group:
  - new: list, rename, modify
  - rearranged: create

- transport:
  - new: list
  - rearranged: format, mount, unmount

- still in the top group:
  - active, status, sync, import-files (could be moved into a new subgroup)

TBD:
- should we have commands for listing and querying acquisitions and/or archives? (import-files would then be moved into it)
- should there also be "info" to see the detailed results of each node (`list` shows all nodes but with only a subset of properties)


cubranic:
Example outputs, using the FRB database:
```
$ alpenhorn group list
Name          Notes
------------  -------
frb-archiver
transport
frb-baseband
cedar
```

```
$ alpenhorn node list
Name          Group         Type    Host                    Root                          Notes
------------  ------------  ------  ----------------------  ----------------------------  -------
frb-backup    frb-archiver  F       frb-analysis5           /data/archive/backup
frb-archiver  frb-archiver  F       frb-analysis5           /data/frb-archiver
CH-ZAD7JT3Q   transport     T       frb-analysis5           /mnt/CH-ZAD7JT3Q
CH-ZAD7JVXH   transport     T       frb-analysis5           /mnt/CH-ZAD7JVXH
CH-ZAD7JDNG   transport     T       frb-analysis5           /mnt/CH-ZAD7JDNG
CH-ZAD7JE5D   transport     T       frb-analysis5           /mnt/CH-ZAD7JE5D
CH-ZAD7GCH7   transport     T       frb-analysis5           /mnt/CH-ZAD7GCH7
CH-ZAD7JEZN   transport     T       frb-analysis5           /mnt/CH-ZAD7JEZN
CH-ZAD7JF2D   transport     T       frb-analysis5           /mnt/CH-ZAD7JF2D
CH-ZAD7JWX8   transport     T       frb-analysis5           /mnt/CH-ZAD7JWX8
CH-ZAD7JFVE   transport     T       frb-analysis5           /mnt/CH-ZAD7JFVE
CH-ZAD7JH2F   transport     T       frb-analysis5           /mnt/CH-ZAD7JH2F
frb-baseband  frb-baseband  A       frb-baseband            /data/archive/baseband
cedar         cedar         A       cedar.computecanada.ca  /project/6004902/archive/FRB
```

```
$ alpenhorn transport list
Name         Mounted    Host           Root              Notes
-----------  ---------  -------------  ----------------  -------
CH-ZAD7JT3Q  -          frb-analysis5  /mnt/CH-ZAD7JT3Q
CH-ZAD7JVXH  -          frb-analysis5  /mnt/CH-ZAD7JVXH
CH-ZAD7JDNG  -          frb-analysis5  /mnt/CH-ZAD7JDNG
CH-ZAD7JE5D  -          frb-analysis5  /mnt/CH-ZAD7JE5D
CH-ZAD7GCH7  -          frb-analysis5  /mnt/CH-ZAD7GCH7
CH-ZAD7JEZN  Y          frb-analysis5  /mnt/CH-ZAD7JEZN
CH-ZAD7JF2D  Y          frb-analysis5  /mnt/CH-ZAD7JF2D
CH-ZAD7JWX8  Y          frb-analysis5  /mnt/CH-ZAD7JWX8
CH-ZAD7JFVE  Y          frb-analysis5  /mnt/CH-ZAD7JFVE
CH-ZAD7JH2F  Y          frb-analysis5  /mnt/CH-ZAD7JH2F
```

ketiltrout:
>     * still in the top group:
>       * active, status, sync, import-files (could be moved into a new subgroup)

I'd say all of these belong under `node` which seems like the primary noun associated with all of them. Either as a list type (`active`, `status`) or else as the instanced target which the verb updates (`sync` and `import-files`).

cubranic:
I started off with "active" as a subcommand of 'node", but kept it top-level for now because it lists nodes on a host, so I wasn't sure if maybe _host_ is the target.

Similarly, with "status", because it joins nodes and archives, the target and focus is also a bit less clear-cut.

"sync" was causing trouble to @jrs65 in #122 also, because it has both node and group as a target. I left it as top-level until that was clearer and also because it's already used in existing data transfer automation, so this would be backward-compatible.

"import-files" is the most clear-cut about the target, which is the node. I kept it top-level partly out of hope that there will be room for it in another subgroup because at current pace we are just going to end up with most of the current commands as subcommands of "node".

cubranic:
I'd like to merge this soon -- can we decide what to do on the last few top-level commands? (active, status, sync, import-files)

Don [thinks](#issuecomment-567754821) they should all be subcommands of `node`. I'd like to avoid putting too much into `node`, although it's kind of a natural home for a lot of commands. How about `node active`, and leave the other three as top level commands for now?

ketiltrout:
Didn't you change `import-files` to `node scan` already?

For the others:
* I think I've been convinced that `status` should stay top-level.
* `node active` seems fine for me.
* `sync` is the tricky one.  I'm can't decide if it's better under `group` (the destination), rather than `node` (the source), so maybe leave it top-level for now.  Once we get a feel for how these groups work in practice, maybe it'll become more apparent where it should be.

cubranic:
Ach, you're right: `import-files` is `node scan` now. I'll merge this, and we can fine-tune as we go ahead.

PR #127, opened 2020-03-02
More efficient database querying when verifying only a subset of acqs

cubranic:
When an `--acq` option is specified in the `node verify` command, we can
dramatically reduce the initialy database query by including a `WHERE IN`
clause. Currently, the `--acq` filter is handled after the query, while
iterating through the query's resultset.


cubranic:
Well darn: this otherwise reasonable change only produced a marginal result: running a query on our ginormous FRB database went from 30s to 28s, even though the result was empty set. It appears that the reason was that we didn't have an index on ArchiveAcq.name. When I added it, the query was evaluated instantly.

I feel like we should we add that to the model. How come we haven't yet? I don't see any mention of key indices in our code.

PR #130, opened 2020-03-06
Add a 'acq' subcommand to the CLI for listing details of acquisitions

cubranic:
Current commands:

- `list [NODE]`: with NODE given, lists names of acquisitions and a count of
   their files for data present on the node. Without the argument, lists all
   known acquisitions. (Can be long!)

- `files ACQ`: lists the names of all files that are a part of the acquisition.

### DIFF
--- a/alpenhorn/client/__init__.py
+++ b/alpenhorn/client/__init__.py
@@ -15,6 +15,7 @@ import alpenhorn.storage as st
 import alpenhorn.acquisition as ac
 
 from .connect_db import config_connect
+from . import acq
 from . import group
 from . import node
 from . import transport
@@ -285,6 +286,7 @@ def status(all):
     print(tabulate.tabulate(data, headers=headers, floatfmt=".1f"))
 
 
+cli.add_command(acq.cli, "acq")
 cli.add_command(group.cli, "group")
 cli.add_command(node.cli, "node")
 cli.add_command(transport.cli, "transport")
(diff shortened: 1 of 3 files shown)


jrs65:
How wedded to a noun-verb system are we? I note that files is not a verb. Maybe `alpenhorn acq listfiles`? Or `alpenhorn file list --acq=acqname`?

cubranic:
> How wedded to a noun-verb system are we? I note that files is not a verb. Maybe `alpenhorn acq listfiles`? Or `alpenhorn file list --acq=acqname`? I think the former.

I'm OK with bending the rule in the right circumstances. I think I like `acq files` better than `acq listfiles` or `file list`. (We don't currently have a `file` subcommand, only `acq`, and manipulating "files" is not directly surfaced anywhere else -- IOW, we only select by `--acq`, and never by `--file`.)

Alternatively, we could make this a part of `acq list` and have it be controlled by an option. Either the generic `-v/--verbose`, or the specific `--files`.

cubranic:
@jrs65 any comments on my proposal? Options:

1. leave `acq files`, even though it breaks the NOUN-VERB pattern
2. make this a part of `acq list`, controlled by an option. Either
  a. `--files`
  b. `-v/--verbose`

3. (least favourite): introduce the `files` subcommand, and rename `acq files` to `files list`.

jrs65:
I don't think we need to be bound by the noun-verb pattern. I think `files` is more obvious what it's doing so let's just go with your option 1.

PR #132, opened 2020-05-07
Fix a failing unit test caused by checking for wrong quotes

cubranic:
This definitely did not use to fail, and it check an error message thrown by Click itself, so it might be a recent update switched between single and double quotes to show valid choices for a multi-choice parameter.


cubranic:
Or should we use a regular expression and be able to match on either quote here, Richard?

PR #135, opened 2021-04-20
Modernise code base

jrs65:
@cubranic these are the changes I've made

- remove Python 2 support from code
- apply black formatting to code base
- raise minimum Python version to 3.6
- remove * imports and sort into the canonical order
- docker tests crash even if not being run
- move to Github actions for CI


jrs65:
Obviously there's a lot of lines changed, but if you just look through each commit separately it should be pretty efficient to review.

tests/docker/test_docker.py
```diff
@@ -463,9 +463,11 @@ def test_sync_acq(workers, network, test_files):
 
 
 def _verify_clean(acqs, worker, unclean=False, check_empty=False):
-    """Check the clean command has been executed as expected on the node associated with 'worker'.
-    If 'unclean' is set to True, check that files are not wanted but still present (until
-    additional copies on other archive nodes are found).
+    """Test the clean command.
+
+    Chek the comand has been executed as expected on the node associated with
```
cubranic, line 468: "Chek" spelling

alpenhorn/client/__init__.py
```diff
@@ -21,7 +21,6 @@
 def cli():
     """Client interface for alpenhorn. Use to request transfers, mount drives,
     check status etc."""
-    pass
```
cubranic, line 31: Can you have empty method body like this? I always thought "pass" was required here.

tests/test_client.py
```diff
@@ -11,9 +11,9 @@
 from click.testing import CliRunner
 
 try:
-    from unittest.mock import call, patch
+    pass
```
cubranic, line 14: This doesn't look right! I think the old construct imported `call` and `patch` mocks from the package that was named differently in py2 vs py3. This does nothing in `try` and then does nothing again if that fails. :-) The package name in py3 is `unittest.mock`, so shouldn't we just keep the one statement, `from unittest.mock import call, patch`?

cubranic: Or, maybe I'm missing something and the import was not needed any more (since CI tests pass?), in which case we can remove this whole try-except segment.

cubranic (approved):
👍 

So what's with the Docker tests? They only fail in CI or standalone too?

cubranic:
A question about the CI: does
```
on:
   pull_request:
     branches:
     - master
   push:
     branches:
     - master
```
mean it will run on any PR against master as well as on master itself after the PR has been merged?

I've run into projects where CI seemed to be set ups to run the same set of actions twice in PRs, which seemed really annoying and wasteful, although I never really spent time to look into what's causing it, as I was only a peripheral contributor.

alpenhorn/client/__init__.py
jrs65: Yeah, I think if you have a doc string it's fine as it counts as part of a function body.

tests/test_client.py
jrs65: Yeah, indeed. It seems it's not needed at all any more.

jrs65:
Yeah, I think it does. Although I think it skips it if it you do a rebase or it can do a fast-forward merge.

I agree it is annoying particularly for kotekan, although I think it makes some sense. The HEAD of the PR branch isn't necessarily what the new master will be.

PR #137, opened 2021-07-28
Asynchronous transfers with a task queue and Nearline support

james-s-willis:


james-s-willis:
@jrs65 @ketiltrout I've added the nearline test, so this is ready to review now

anjakefala:
From a conversation with James.

This PR has the following goals:
* To add an API with which to make file copy requests from nearline tapes. The following documentation was referenced for using lfs to query tape storage: https://docs.computecanada.ca/wiki/Using_nearline_storage
* In order to use nearline, calls now need to be async (nearline has much longer waits than `alpenhorn` has been previously accustomed to). To accomodate this a Task Queue was added, with a threadpool, whose tasks will then be processed.
* To add tests for the new API. `tests/lfs` contains a mock for `lfs`.
* To begin bringing over fixes that have been added to `alpenhorn 1`, but have not been ported to `alpenhorn 2`, yet (e.g. https://github.com/chime-experiment/alpenhorn/pull/19).

`task.py` contains the new Task logic. `update.py` has been then factored out, and can now use `add_task(TaskClass)` for every type of `Task` type.

The build is currently red because some tests check for help output, and one Python package has been updated since the code was written, changing the output. It still needs to be decided whether to fix the version of the Python package or update the tests.

alpenhorn/Task.py
```diff
@@ -0,0 +1,1021 @@
...
+            threads.append(threading.Thread(target=self.run_tasks, daemon=True))
+            threads[t].start()
+
+
+class Task:
+    def __init__(self, node):
+        self.node = node
+
+    def run(self):
+        print("{} run()...".format(type(self).__name__))
```
anjakefala, line 61: Instead of being a print statement, this should probably be a `raise NotImplementedError`

alpenhorn/Task.py
```diff
@@ -0,0 +1,1021 @@
...
+                if util.md5sum_file(fullpath) == fcopy.file.md5sum:
+                    log.info("File is A-OK!")
+                    fcopy.has_file = "Y"
+                    copy_size_b = os.stat(fullpath).st_blocks * 512
+                    fcopy.size_b = copy_size_b
+                else:
+                    log.error("File is corrupted!")
+                    fcopy.has_file = "X"
+            else:
+                log.error("File does not exist!")
```
anjakefala, line 99: Let's document the fullpaths in these error logs. Because now there might be multiple threads going, and the logs might get interleaved.

alpenhorn/Task.py
```diff
@@ -0,0 +1,1021 @@
...
+        """Process file copy requests onto this node."""
+
+        print("{} run()...".format(type(self).__name__))
+
+        global done_transport_this_cycle
+
+        avail_gb = self.node.avail_gb
+
+        # Skip if node is too full
+        if avail_gb < (self.node.min_avail_gb + 10):
```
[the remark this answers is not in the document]
anjakefala: I guess that gets calculated later!

alpenhorn/Task.py
```diff
@@ -0,0 +1,1021 @@
...
+
+class Task:
+    def __init__(self, node):
+        self.node = node
+
+    def run(self):
+        print("{} run()...".format(type(self).__name__))
+
+
+class IntegrityTask(Task):
```
anjakefala, line 64: James factored this out of `update.py`, he did not write the logic in `IntegrityTask`.

alpenhorn/Task.py
```diff
@@ -0,0 +1,1021 @@
...
+            else:
+                log.error("File does not exist!")
+                fcopy.has_file = "N"
+
+            # Update the copy status
+            log.info("Updating file copy status [id=%i]." % fcopy.id)
+            fcopy.save()
+
+
+class DeletionTask(Task):
```
anjakefala, line 107: This is a new task.

alpenhorn/util.py
```diff
@@ -43,7 +43,11 @@ def run_command(cmd, **kwargs):
     stdout_val, stderr_val = proc.communicate()
     retval = proc.returncode
 
-    return retval, stdout_val, stderr_val
+    return (
+        retval,
+        stdout_val.decode(errors="replace"),
+        stderr_val.decode(errors="replace"),
+    )
```
anjakefala, line 50: Why did this get added? Does this affect stdout for other commands written before James' code?

alpenhorn/Task.py
```diff
@@ -0,0 +1,1021 @@
...
+                log.info("Too few backups to delete %s" % shortname)
+
+
+class DiskTransferTask(Task):
+    def run(self):
+        """Process file copy requests onto this node."""
+
+        print("{} run()...".format(type(self).__name__))
+
+        global done_transport_this_cycle
```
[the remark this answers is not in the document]
anjakefala: I do not think it does. At minimum, we will need to import `done_transport_this_cycle`. But even with that, I am not totally sure if the `global` keyword allows the setting of globals from another file, I am going to run a test.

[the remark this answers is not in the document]
anjakefala: But even if it works, we are probably better off doing something else than this.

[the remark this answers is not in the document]
anjakefala: (The `global` keyword does allow the setting of globals from another file.)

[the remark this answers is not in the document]
jrs65: Yup. 

I think done_transport_this_cycle is only really necessary on the DiskTransfer tasks, so it could be moved to a class variable instead. That's definitely a lot cleaner.

There might still be a better way of doing this though. I'm always a little unclear of what the point of it is. I *think* it's to make sure that transport disk copies end up filling up each disk in turn rather than spreading out over all the available disks.

alpenhorn/Task.py
```diff
@@ -0,0 +1,749 @@
...
+
+            else:
+                log.info("Too few backups to delete %s on %s" % (shortname, self.node.name))
+
+class TransferTask(Task):
+
+    def __init__(self, node):
+        # how do the double inits work?
+        super().__init__(node)
+        self.fs_type = None
```
anjakefala, line 224: Let's set these as class variables, instead of in the `__init__`.

ketiltrout:
Whilst the functional parts of James's work is still present, I've largely rewritten the task/queue framework.

I'll going to create a new PR for that.

PR #144, opened 2023-01-28
Rewrite 1/10: preliminaries and CHIMEDB integration

ketiltrout:
* _This is the first of ten PRs for the rewrite of alpenhorn._

## Overview

This PR starts out by doing some maintenance, and then does updates `db.py` and `extensions.py`  to integrate with the new `chimedb.core.alpenhorn` module I've introduced over in https://github.com/chime-experiment/chimedb/pull/34.

## Housekeeping

The housekeeping work done to prime this re-write is as follows:
 * minimum Python version has been raised from 3.6 to 3.9.  The CI workflow is now run on version ~3.9~ 3.10 and 3.11 instead of 3.6 and ~3.9~ 3.10.
 * I have deleted the `/legacy/` directory which contained some of the code from the original `alpenhorn-1` implementation
 * I have deleted all of the YAML-based test data which was in the `tests/fixtures` directory.  This has broken the marjority of existsing tests, which I've simply disabled for now.  As more parts of the code gets re-written, I'll be introducing fixes to the test suite to cover the new code.
 * I've also moved all the client tests to a `tests/client` subdirectory to mirror the organisation of the code itself.
 * Running `black` has caused a reflow of text all over the place.

## Changes to what a database extension provides

The essential change here is that instead the `register_extension` function of a database extension returning an function
that initialises `peewee.Database`, it instead returns a dict of "capabilities", which includes a key called "connect" key that provides the `peewee.Database`-making function.

The capability dict also contains a `close` function to close the database (probably unnecessary) and a boolean indicating whether the database extension is threadsafe or not.

If no database extension module is provided, or one or more of these keys is missing from the capability dict, then alpenhorn's `db.py` provides fallback implementations of all of these.  The fallback is _not_ threadsafe.

(Implementation note: I originally had everything that `chimedb.core` could possibly provide in this capability dict, but as the alpenhorn code matured, I came to the conclusion that more and more of it was more trouble than it was worth, and I've now pared it down to this minimal set.)

The function `extensions.connect_database_extension` has largely been replaced by `db.init` which sets-up the `db` module based on whatever database extension was found (if any).  The `extensions` module also now complains if more than one database extension is specified in the config (since only one of them could possibly be used).

Operationally, the changes here mean that alpenhorn start-up now needs to have a
```
db.init()
```
call after extensions are loaded but before the first `db.connect()` call to set up the `db` module. Additionally, each thread that needs to access the database also needs to call `db.connect()` before doing that.

## Peewee-3 changes

I have ported the fixes made in `chimedb.core` to both the mixin `RetryOperationalError` and the `EnumField` class so that they work in peewee-3.  See https://github.com/chime-experiment/chimedb/pull/34 for details.  In practice, when using the CHIME database extension, we don't use the alpenhorn-provided `RetryOperationalError`.  We _do_ use the `EnumField` defined here, but don't need the fix because MySQL has a native Enum type making the fix unnecessary.


ketiltrout:
I realised I can't make stacked PRs if they're off in some other fork, so I've remade #143 locally.

tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+    # Reset globals
+    config.config = None
+    extensions._db_ext = None
+    extensions._ext = None
+
+
+@pytest.fixture
+def use_chimedb(set_config):
+    """Use chimedb, if possible.
+
```
jrs65, line 44: It might be worth pointing out that this doesn't actually run against the live chimedbrc.

tests/test_db.py
```diff
@@ -1,64 +1,92 @@
...
+
+    # Test table
+    class Values(db.base_model):
+        value = pw.IntegerField()
+
+    # Create and populate
+    dbproxy.create_tables([Values])
+    assert dbproxy.get_tables() == ["values"]
+    Values.insert_many([{"value": 12}, {"value": 34}, {"value": 56}]).execute()
+
```
jrs65, line 33: For understanding the exact tests I think it might be useful to test the db contents at this point.
```suggestion

     assert list(Values.select().dicts()) == [
         {"id": 1, "value": 12},
         {"id": 2, "value": 34},
         {"id": 3, "value": 56},
     ]
     
```

jrs65 (changes requested):

alpenhorn/db.py
```diff
@@ -1,83 +1,167 @@
...
+            Invoked to close the database connection.
+
+Before access the attributes of this module, init() must be called to set
+up the database.  After that function is called, the following attributes
+are available:
+
+- threadsafe: a boolean indicating whether the database can be concurrently
+            accessed from multiple threads.
+
+Also after calling init(), a connection to the database may be initialised
```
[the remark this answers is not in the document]
ketiltrout: Yes, good point.

alpenhorn/extensions.py
```diff
@@ -131,6 +134,7 @@ def _register_acq_extensions(acq_types):
 
 
 def _register_file_extensions(file_types):
+    from . import acquisition
```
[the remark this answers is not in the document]
ketiltrout: Yes!  It prevents an import loop:
```
alpenhorn/db.py:41: in <module>
    from . import config, extensions
alpenhorn/extensions.py:27: in <module>
    from . import acquisition
alpenhorn/acquisition.py:8: in <module>
    from .db import base_model
E   ImportError: cannot import name 'base_model' from partially initialized module 'alpenhorn.db' (most likely due to a circular import) (/Users/dvw/git/rc_alpenhorn/alpenhorn/db.py)
```
(which is what happens if I move this import to the top here.)

This is a temporary fix.  A better solution will appear in PR 6/10.

ketiltrout:
Mostly doc updates and type hinting.  I've updated the minimum version to 3.10 for type hinting. 

I also found some more unittests in `tests/test_extensions.py` that were supposed to be in this PR but got overlooked the first time round.

alpenhorn/db.py
```diff
@@ -87,20 +190,32 @@ def _connect(url=None, db=None):
 class RetryOperationalError(object):
```
jrs65, line 190: Not a line changed by you, but the `(object)` isn't need in python 3.
```suggestion
class RetryOperationalError:
```

jrs65 (approved):

PR #145, opened 2023-01-28
Rewrite 2/10: The Fair Multi-FIFO Queue

ketiltrout:
This PR introduces a "Fair Multi-FIFO Queue", which will be used by alpenhorn as the task scheduler.  (It doesn't use it yet, as the code introduced here isn't actually used anywhere yet.

## Overview

The FMFQ is a multi-producer, multi-consumer queue designed to be similar to the standard `queue.Queue` though it doesn't derive from that code, nor is it a drop-in replacement for it.  All items in the queue are assigned to a FIFO (of which there are many).  The FIFO for a given item is specified by a key.  The queue is "fair" in the sense that it tries to keep the same number of items in-progress at any given time.  (Same number _in-progress_ not _queued_.)

The FIFOs are implemented as `deque`s, though I don't know if that really saves us a whole lot of time.

The queue is unbounded.

## Queue basics

Like the standard `queue.Queue`, there are three primary functions used to interact with the queue:
* `put(item, key, wait=0)` adds `item` to the queue, putting it at the end of the FIFO named `key`.  (For `wait`, see "Deferred puts" below).
* `get(timeout=None)` retrieves the next item from the queue (where the concept of "next" adheres to the fairness criterion described above).  If there is currently nothing in the queue, this function either waits for ever or else until `timeout` expires.   When returning an item from the queue (i.e. not timing out), it also provides the `key` to indicate which particular FIFO the item came from. When this function provides an item from the queue, that item is marked as being _in-progress_.
* `task_done(key)` tells the queue that an item that was obtained from the FIFO named `key` is no longer in-progress.  A typical consumer will have code that looks something like this:
```
item, key = queue.get()
# Do something with item
queue.key(key)
```
The queue doesn't actually care which items are in-progress.  It only keeps track of the number of in-progress items per FIFO. 

## Other methods

There is a `join` method used to shutdown the queue.  It basically waits until everything from the queue is finished.  It's used at the end of alpenhorn when it attempts to shutdown cleanly.

There are also a number of queue-size methods which will be used by alpenhorn to determine whether a node is "idle" or not.

## How the queue is used in alpenhorn

In alpenhorn, queued items are I/O tasks (which will be introduced in a subsequent PR), but that doesn't matter to the queue itself.  The queue can accept any object.  In alpenhorn, the FIFO `key`s are node names, meaning there is a separate queue per node.  The fairness of the queue is needed to prevent one node (i.e. Nearline) from monopolising the worker threads.  If only nearline tasks are present in the queue, they'll be provided to consumers, but if tasks from other nodes are also present, nearline jobs won't overwhelm the system and prevent these other tasks from being handled in a timely manner.

While the queue code doesn't have anything alpenhorn-specific, a few design decisions have been made based on how alpenhorn uses the queue.  Specifically, we assume the number of distinct FIFOs is small and even if they empty out, they will be re-used with the same key.  The implementation here also assumes the number of items put into the queue is large in comparison to the number of FIFOs.  (See the overly verbose comments in the code for further details.)

## Deferred puts

If `put()` is called with a positive value for `wait`, then the put is accepted by the FMFQ but _not_ added to the queue.  Instead it gets put into a `heapq` of "deferred puts".  These items wait around for `wait` seconds to elapse before the item is added to the queue.

Alpenhorn uses deferred puts to implement tasks which need to wait around for something to happen (i.e. nearline recalls).  Having these tasks wait in the "deferred put" heap, means they aren't available for workers to take and the nodes appear idle until the deferrals expire.

The deferred puts are resolved during a `get` call (which will check for expired deferrals).  As a result, often deferred puts never actually show up in the queue, since they're immediately `get`ted.

## Other changes
I've added a workaround for the DB concurrency test.  I don't understand why it fails when using the Sqlite3 connector.  I've also added the CHIMEDB core package to the test-requirements so we can test it with this code.

### DIFF
--- a/alpenhorn/queue.py
+++ b/alpenhorn/queue.py
@@ -1,0 +1,420 @@
+"""Fair Multi-FIFO Queue
+
+The Fair Multi-FIFO Queue implements a multi-producer, multi-consumer queue
+similar to the queues implemented by the queue module.  This queue has all
+the same locking semantics and thread-safety as queue.Queue but is not
+subclassed from that.
+
+Each task in the Fair Multi-FIFO Queue is part of a FIFO specified by a key.
+The Fair Multi-FIFO Queue may have many such FIFOs, and it will try to ensure
+that tasks are removed from the queue in a fair manner which tries to keep
+the same number of tasks from each FIFO in progress at all times.
+
+The queue is unbounded.
+"""
+
+import heapq
+import threading
+from time import time
+from typing import Any
+from collections import deque
+from collections.abc import Hashable
+
+
+class FairMultiFIFOQueue:
+    """Create a new Fair Multi-FIFO Queue"""
+
+    __slots__ = [
+        "_all_tasks_done",
+        "_deferrals",
+        "_dlock",
+        "_fifos",
+        "_inprogress_counts",
+        "_joining",
+        "_keys_by_inprogress",
+        "_lock",
+        "_not_empty",
+        "_total_inprogress",
+        "_total_queued",
+    ]
+
+    def __init__(self) -> None:
+        # The FIFO dict
+        self._fifos = {}
+
+        # Total number of queued tasks
+        self._total_queued = 0
+        # Total number of in-progress tasks
+        self._total_inprogress = 0
+        # Counts of in-progress tasks by FIFO
+        self._inprogress_counts = {}
+        # A list of sets of FIFO keys, indexed by number of in-progress tasks
+        #
+        # We initialise element 0 to the empty set to simplify creating new
+        # FIFOs, which will always get added to that set.
+        self._keys_by_inprogress = [set()]
+
+        # The thread lock (mutex) and the conditionals (see queue.py for
+        # details, which this implementation broadly follows)
+        self._lock = threading.Lock()
+        self._not_empty = threading.Condition(self._lock)
+        self._all_tasks_done = threading.Condition(self._lock)
+
+        # Deferred puts.  This is a heapq.
+        self._deferrals = list()
+        # Are we in a join() call?
+        self._joining = False
+        # The lock for _deferrals and _joining, which can be held independently
+        # of the primary _lock.
+        self._dlock = threading.Lock()
+
+    def task_done(self, key: Hashable) -> None:
+        """Report that a task from the FIFO named `key` is done.
+
+        After a consumer has finished processing a task provided to it
+        by a get(), the key provided by that get() must be passed back
+        into this function to report the completion of that task.
+
+        Parameters
+        ----------
+        key : hashable
+            The name of the FIFO that the completed task came from.  The
+            value passed in should previously have been returned from a
+            `get()` call.
+
+        Raises
+        ------
+        ValueError
+            The FIFO named `key` has no in-progress tasks.
+        """
+
+        with self._all_tasks_done:
+            # The number of currently in-progress tasks (including the one
+            # that's finishing now)
+            count = self._inprogress_counts.get(key, 0)
+
+            # If the specified FIFO has no in-progress tasks, raise value error
+            if count <= 0:
+                raise ValueError(f"no unfinished tasks for FIFO {key}")
+
+            # Remove the FIFO from the ordered list
+            self._keys_by_inprogress[count].remove(key)
+
+            # Decrement counts
+            count -= 1
+            self._inprogress_counts[key] = count
+            self._total_inprogress -= 1
+
+            # Put the fifo back in the ordered list.
+            #
+            # XXX Could just delete a FIFO which is now empty here.
+            #
+            # In general, there's no reason to expect an empty FIFO to be reused,
+            # by the caller, and numerous empty FIFOs will slow down get() over time,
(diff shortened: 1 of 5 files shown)


alpenhorn/queue.py
```diff
@@ -0,0 +1,426 @@
...
+        if key not in self._fifos:
+            fifo = self._new_fifo(key)
+        else:
+            fifo = self._fifos[key]
+
+        # push the task onto the FIFO (right-most end)
+        fifo.append(item)
+        self._total_queued += 1
+
+    def put(self, item: Any, key: Hashable, wait: int = 0) -> None:
```
ketiltrout, line 248:
```suggestion
    def put(self, item: Any, key: Hashable, wait: float = 0) -> None:
```

alpenhorn/queue.py
```diff
@@ -0,0 +1,426 @@
...
+            # Put the fifo back in the ordered list.
+            #
+            # XXX Could just delete a FIFO which is now empty here.
+            #
+            # In general, there's no reason to expect an empty FIFO to be reused,
+            # by the caller, and numerous empty FIFOs will slow down get() over time,
+            # so triming could decreate get() execution time.
+            #
+            # However, in our particular case (alpenhornd), the maximum number
+            # of FIFOs is sum of nodes and groups ever active on this host, which is
```
ketiltrout, line 116:
```suggestion
            # of FIFOs is sum of nodes ever active on this host, which is
```

alpenhorn/queue.py
```diff
@@ -0,0 +1,426 @@
...
+        Parameters
+        ----------
+        key : hashable
+            The name of the FIFO to create.
+
+        Returns
+        -------
+        fifo : deque
+            The newly created FIFO
+        """
```
ljgray, line 217: Although this _is_ a private method, I'm kind of of the opinion that the logic for checking if a FIFO key already exists should be done here rather than in the `_put` method. A user could technically overwrite an existing queue by accessing this directly, and there wouldn't be any explicit warning about it. I'm also not sure if the logic of creating a new FIFO makes sense in the `_put` method (even though that is also private). 

My suggestion is to replace the `_new_fifo` method with something like a `_get_fifo` method which would return either an existing queue or create a new one as needed.

alpenhorn/queue.py
```diff
@@ -0,0 +1,426 @@
...
+        key : hashable
+            The name of the FIFO to which `item` is added
+        wait : float, optional
+            The number of seconds to wait before adding `item` to the queue.
+        """
+
+        if wait > 0:
+            # Deferred put
+            with self._dlock:
+                # If joining, silently discard this put
```
ljgray, line 273: I'm not convinced that this should be entirely silent, there should probably be some sort of warning that this put will never execute.

alpenhorn/queue.py
```diff
@@ -0,0 +1,426 @@
...
+                while len(key_set) > 0:
+                    key = key_set.pop()
+                    if len(self._fifos[key]) == 0:
+                        empties.add(key)
+                        key = None
+                    else:
+                        break
+
+                # Put all the empty FIFOs back
+                key_set.update(empties)
```
ljgray, line 353:
```suggestion
                # Pull FIFOs out of the set until we find a non-empty one
                for k in key_set:
                    if len(self._fifos[k]) > 0:
                        key = k
                        break
```
I think this could just be a for loop which grabs the first non-empty fifo instead of discarding and re-updating with empty queues.

ljgray (commented):
Looks good overall, I added a few comments to consider (and one string format fix)

alpenhorn/queue.py
ketiltrout: I'm not really sure why this is separate.  What would you think about just merging the four operative lines down into the `_put` function at the only place where this is called?

alpenhorn/queue.py
ketiltrout: Maybe, though I don't think alpenhornd ever actually calls this method, because if it did, it would be the last thing it does before crashing.  I think I mostly wrote this for completeness sake and/or unit tests.

In alpenhorn, deferred puts can only happen as the result of a `task` that needs to wait for something to happen.  So, even if alpenhorn were to call this message, I don't think it would do it at a time when tasks are still ongoing (because the worker pool is shutdown before the task queue).

ketiltrout: At minimum, it can't hurt to change the return type of `put` from `None` to `bool` and indicate the failed put that way.

alpenhorn/queue.py
ljgray: I think that would make sense, and it would remove the possibility of any issues happening (however unlikely). If the method isn't used anywhere else than it probably doesn't have to be its own method

ketiltrout:
I think I've addressed all the comments.

I've also added a second commit that's not directly relevant to these changes, but it's convenient to add here.  It adds `chimedb` to the test requirements (and also fixes a concurrency test that's failing due to something in the Sqlite3 connector which I don't get.)

ljgray (approved):
Great!

PR #147, opened 2023-01-30
Rewrite 4/10: Table model updates

### LINKED ISSUE #49: What is the purpose of `storage_type` "F"?
The docstring says "'F': for data in the field (i.e acquisition machines)", but I don't see it being checked for anywhere in the code.

ketiltrout:
This PR updates the peewee table models for the data index (`StorageGroup`, `StorageNode`, `ArchiveAcq`, `ArchiveFile`, `ArchiveFileCopy`, `ArchiveFileCopyRequest`).  It does _not_ deal with `AcqType` or `FileType`, which will be handled in a later PR.

This PR updates the field lists for these tables (adding new fields and removing unused fields).  All field updates are backwards compatible with alpenhorn-1.  It also add a few convenience methods for various common database queries related to the tables.  The methods are mostly database queries that were performed explicitly within the I/O code.  Moving these queries to methods here is to reduce the complexity of the I/O code itself.

`StorageGroup` and `StorageNode` will have a subsequent update to add the new I/O framework, but the other table models (`ArchiveAcq`, `ArchiveFile`, `ArchiveFileCopy`, `ArchiveFileCopyRequest`) have no further changes pending.

## StorageGroup
Fairly light changes:
* Marks the `name` field as unique
* Adds a `copy_state` method which takes an `ArchiveFile` and returns the value of `has_file` (Y/M/X/N) for copy/copies of the `ArchiveFile` in this group.

## StorageNode
A more extensive update.
### StorageNode fields
* `name` is marked unique
* `suspect` is dropped.  It was never used or implemented properly.
* No change to `storage_type`, but I have updated its description in the docstring to better indicate what 'F' means (i.e. it's anything that's not an archive and not transiting storage).  This closes #49 .
* `max_total_gb` allows null/None (for nodes which can't determine their maximum size).  The unusual default of -1 is dropped. The code retains support for using -1 to indicate an unknown `max_total_gb`, but will never set the value to -1 (it will use null/None instead).
* `min_avail_gb` default explicitly set to zero.
* `min_delete_age_days` is dropped.  It has never been used.
### StorageNode methods and properties
* `local`: a boolean property which is True when this StorageNode is local to the running alpenhorn (whether or not it is active).
* `archive`: a boolean property which is True when the StorageNode is an archive node (when `storage_type=='A'`)
* `under_min`: Returns False if `avail_gb` is None, otherwise returns `min_avail_gb > avail_gb`.
* `over_max`: Returns False if `max_total_gb` is None or non-positive, or else a boolean: sum of all `ArchiveFile.size_b` on this node (see `total_gb` below) is greater than `max_total_gb`.
* `named_copy_present` given an acq name and a file name, returns True if a copy of the file in the acq exists on this node.
* `total_gb` sums up `ArchiveFile.size_b` for all files on this node and returns the result in GiB.
* `all_files` returns a list of paths (relative to `root`) for all files existing on this node.

## ArchiveAcq
Only change:
* `name` is marked unique.

## ArchiveFile
Field changes:
* adds a constraint that the tuple (`name`, `acq`) is unique.  This constraint is already present in the CHIMEDB production database.  We may have added it by hand there.

Properties and methods:
* `path` the path to this file, i.e. the path concatenation of `acq.name` and `name`.
* `archive_count` returns the number of archive nodes containing a copy of this file.

## ArchiveFileCopy
Fields:
* Add `ready`, a boolean which _might_ indicate that a file is ready for I/O.  This field is used to communicate between hosts which files are available for pulls, but, as the docstring indicates, shouldn't ever be consulted directly, because different I/O types can have different interpretations of the meaning of this field.  (In James's work, we called this "prepared").
* Allows `size_b` to be null/None (for instances where the size-on-node can't be determined).
* Adds `last_update` a timestamp that is automatically updated whenever the `ArchiveFileCopy` record is changed.

Properties and Methods:
* To implement the automatic `last_updated`, the `save` and `update` methods are re-implemented to add the new `last_udpated` value before handing things off to peewee.  This could also be done at the DB level: MySQL allows an `ON UPDATE CURRENT_TIMESTAMP` clause in the table schema, but that isn't supported by peewee.
* `path` a property returning an absolute `pathlib.Path` to the file copy.

## ArchiveFileCopyRequest
Field changes:
* `nice` is dropped.  Has never been used.
* `completed` has a default value of False added, which I think _is_ the default of a non-null boolean field if none is given, but it's better to be explicit.
* `timestamp` is given the current time as a creation-time default

## Other changes
Because it's used in the `StorageNode.local` property, this is a convenient place to update the hostname-finding logic:
The function `alpenhorn.util.get_short_hostname` is renamed to `alpenhorn.util.get_hostname` and now supports specifying the hostname explicitly in the `alpenhorn` YAML config, giving us the flexibility to put logical values in `StorageNode.host` (like, say, "scinet"), instead of being forced to use whatever value the actual name of the host we're running on has (like, say, "nia-dm1").

The unit tests for (at least the updated parts of) `alpenhorn.acqusition`, `alpenhorn.archive`, and `alpenhorn.storage` have been restored/updated after being disabled in #144.  (They were also renamed to drop `_model` from the name of the test files, because I like testing `alpenhorn/<name>.py` in `tests/test_<name>.py`).

A large set of DB-data-producing fixtures has been added to `conftest.py`.  These will be used _extensively_ by the I/O framework unit tests.


alpenhorn/storage.py
```diff
@@ -52,26 +108,129 @@ class StorageNode(base_model):
...
+        """Is the total size of file copies on the node greater than allowed?
+
+        If max_total_gb is non-positive or None, returns False.
+        """
+        if self.max_total_gb is None or self.max_total_gb <= 0:
+            return False
+        return self.total_gb() >= self.max_total_gb
+
+    def named_copy_present(self, acqname: str, filename: str) -> bool:
+        """Is an ArchiveFileCopy named ``acqname/filename`` present?
```
ketiltrout, line 159:
```suggestion
        """Is an ArchiveFileCopy named "`acqname`/`filename`" present?
```

alpenhorn/storage.py
```diff
@@ -52,26 +108,129 @@ class StorageNode(base_model):
...
+                ArchiveFileCopy.node == self,
+                ArchiveFileCopy.has_file == "Y",
+            ).get()
+        except pw.DoesNotExist:
+            return False
+
+        return True
+
+    def copy_present(self, file: ArchiveFile) -> bool:
+        """Is a copy of ArchiveFile file present on this node?"""
```
ketiltrout, line 186:
```suggestion
        """Is a copy of ArchiveFile `file` present on this node?"""
```

alpenhorn/acquisition.py
```diff
@@ -306,8 +303,39 @@ class ArchiveFile(base_model):
     name = pw.CharField(max_length=255)
     size_b = pw.BigIntegerField(null=True)
     md5sum = pw.CharField(null=True, max_length=32)
+    # Note: default here is the now method itself (i.e. "now", not "now()").
+    #       Will be evaulated by peewee at row-creation time.
```
ljgray, line 307:
```suggestion
    #  Will be evaulated by peewee at row-creation time.
```

alpenhorn/acquisition.py
```diff
@@ -306,8 +303,39 @@ class ArchiveFile(base_model):
...
 
+    class Meta:
+        # (acq,name) is unique
+        indexes = ((("acq", "name"), True),)
+
+    @property
+    def path(self) -> pathlib.PurePath:
+        """The relative path to the file copy.
+
+        Simply the path contcatenation of acq.name and name.
```
ljgray, line 318:
```suggestion
        Simply the path concatenation of acq.name and name.
```

alpenhorn/storage.py
```diff
@@ -52,26 +108,129 @@ class StorageNode(base_model):
...
+    @property
+    def local(self) -> bool:
+        """Is this node local to where we are running?"""
+        return self.host == util.get_hostname()
+
+    @property
+    def archive(self) -> bool:
+        """Is this node an archival node?"""
+        return self.storage_type == "A"
+
```
ljgray, line 139: Any of the methods below without arguments could probably be properties, for consistency

ljgray (commented):
A couple of minor comments but overall looks good

jrs65 (changes requested):

alpenhorn/storage.py
ketiltrout: Sure.  I can do that.  `under_min` definitely should be a property.

In Python, is there a concept of "maximum work" before a method should be used instead of a property?  The `all_files` method could be a property but it could potentially generate thousands of `pathlib.Path`s every time it's accessed.  `total_gb` (also called by `over_max`) performs an expensive SQL query.  Are either of those a problem?

alpenhorn/storage.py
```diff
@@ -52,26 +108,129 @@ class StorageNode(base_model):
...
+        Returns
+        -------
+        all_files : list
+            a list of paths relative to `root` for all file copies
+            exsiting on the node.
+        """
+        from .archive import ArchiveFileCopy
+
+        return [
```
ketiltrout, line 229: Should return a set, not a list, because we're going to be doing `path in all_files()` on this output, not iterating through it.

alpenhorn/storage.py
ketiltrout: putting a verb into the name of a method isn't a bad idea.

alpenhorn/archive.py
```diff
@@ -7,39 +9,83 @@
...
+        """
+
+        # Only update if dirty
+        if self.is_dirty():
+            self.last_update = datetime.datetime.now()
+            if only is not None and ArchiveFileCopy.last_update not in only:
+                only.append(ArchiveFileCopy.last_update)
+        return super().save(force_insert, only)
+
+    # Re-implement update() to add last_update if not given
```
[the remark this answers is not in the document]
ketiltrout: Hmmm.   I agree coleifer usually knows what he's talking about, but it would be really nice to have an explanation of _how_ it's a problem.  Is it because we need to deal with the `__data` parameter?

Maybe the best way to do it is in the schema, then.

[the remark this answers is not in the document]
ketiltrout: Or we could just update the field manually.  Let me have a think about how important it is to be correct.  If it's very important we should push it down into the schema so that no matter how you go about updating the table it gets updated.  If it's not that important, it's probably better to be cautious here.

ketiltrout:
I made `ArchiveFile.archive_count` and `StorageNode.under_min` into properties.  I've added verbs to the other parameter-free methods, but I'm willing to revert their names and make them properties, too, if that seems better.

Also, I'm now realising the contents of this PR would probably have been better introduced when they were used in the I/O rewrite, but unlikely to be worth changing now.

alpenhorn/archive.py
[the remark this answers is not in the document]
ketiltrout: Well, `ON UPDATE CURRENT_TIMESTAMP` isn't too difficult, so let's do that.  I thought a row-update timestamp column was a fairly normal thing to have in a relational database. It is kind of disappointing that peewee doesn't really support it.

Doesn't work in SQLite, though.  Need to figure out if that needs some more work.

ketiltrout:
I pushed the timestamp update into the schema.  Only downside is it only works in MySQL.

alpenhorn/archive.py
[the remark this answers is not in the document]
jrs65: Yeah. Much of my fear is that given that he doesn't think it's a good idea that a future peewee change might break what we're doing here. That has bitten us before.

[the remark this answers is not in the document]
jrs65: Apparently you can use a trigger into SQLite to make it work. A bit of a weird, and I think you need to do it explicitly in the db.

jrs65 (approved):
This is great. I definitely appreciate the thorough and high quality tests.

PR #148, opened 2023-02-06
Rewrite 5/14: I/O framework

### LINKED ISSUE #60: alpenhornd ignores new mounted nodes if mounted node count drops to zero
We've seen on `jingle` the following behaviour:
* `alpenhornd` working fine with transport disks mounted
* All transport disks are unmounted, leaving `jingle` with no mounted storage nodes
* New transport disks are added to jingle and mounted using `mount_transport`
* Although the new nodes are now properly mounted on `jingle` in the database, `alpenhornd` ignores them, and continues to operate as if there are no nodes mounted.

An `alpenhornd` restart is required to get out of this state.

ketiltrout:
This PR introduces the I/O framework for `StorageNode`s and `StorageGroup`s, although none of the actual I/O is implemented in this PR.  It also rewrites the daemon main loop to integrate the queue, workers, and new I/O framework, although, again, the existing I/O is left as-is for now and does not make use of the new asynchronous task structure.

## StorageNode and StorageGroup I/O framework
This PR adds two new optional (i.e. null-allowing) string fields to both `StorageGroup` and `StorageNode`:
* `io_class` which provides the name of the I/O class for a node/group.  Node and group I/O classes are separate, and a the list of valid names is different for each.  If this field is left empty, it is treated as if it had the value `Default`.
* `io_config` which provides a JSON blob of configuration data, if any, to be interpreted by the I/O class.  The Default I/O classes ignore this field, but others (like Nearline) will require certain data.

The I/O classes can be found in the `alpenhorn/io` directory, although it's pretty barren for the moment:
*  The base I/O classes, which don't actually do any I/O, are in the file `base.py`.  `"base"` cannot be used for the `io_class` property.  (I mean, the field _can_ be set to that value, but it won't work, because `base.py` and the classes `BaseNodeIO` and `BaseGroupIO` have different capitalisation.)
* The Default I/O class, which is what you get if you leave `io_class` empty, is implemented in `Default.py`.  But for now, it's just a re-class of the classes in `base.py` to make I/O class instantiation work properly in the tests.  The `DefaultNodeIO` class is for a "standard" POSIX filesystem.  The `DefaultGroupIO` class is for a simple StorageGroup with only one node in it.

In general, there is no requirement that a group of a given type be populated with nodes of the same type.  The `DefaultGroupIO` class, for instance, doesn't care at all what the `io_class` of its constituent node is.   (There can be exceptions: `NearlineGroupIO` class needs a `NearlineNodeIO` node in it.) Some I/O classes implement only one of Group or Node I/O.

## UpdateableNode and UpdateableGroup
The I/O class for a Storage object is managed by a new pair of classes defined in `update.py`: `UpdateableNode` and `UpdateableGroup`.  These are primarily container classes for a Storage object and it's I/O class instance.  They provide access to both objects for the update code.

These Updateable instances persist through update loops when possible.  A node or group going away will cause the Updatable instance being destroyed.  The Storage object they contain are replaced ever loop (because new Storage objects are created as a side effect of querying the database for the list of current nodes).  I/O instances are only re-instanced between loops if the Storage object's `id`, `io_config` or `io_class` change.

`UpdateableNode` subsumes the old top-level functions in `update.py` for updating nodes (so, e.g. `update_node()` is now `UpdateableNode.update()`.

## Daemon start-up and main loop rewrite
This PR updates the daemon for the db changes in #144, and ties in the new functionality of the task queue (#145) and the worker pool (#146).

The daemon start-up now has the necessary `db.init()` call, and also now instantiates the queue and worker pool in preparation for multithreading (although there are no actual tasks yet being used in this update).  The initial number of workers to start has been added to the `service` section of the alpenhorn config.

The main loop has been re-written for the new I/O framework. A single iteration of main loop looks like this:

**The I/O update:**
* query the database for the list of active nodes on the host
* (re-)instantiate all the I/O classes for the active nodes
* Loop over nodes:
  * check if the node is actually active on the host (`update_node_active`).  If it isn't forget about this node.
  * check the queue to determine whether this node is idle at the start of the loop, or if it still has on-going I/O from a previous iteration.  (If it isn't idle, it's not going to be updated this time through the loop).
  * build up a list of active groups on the host (by finding unique values of `node.group`).
  * Call the I/O layer `before_update` hook.  This function may also cancel the node update for this time through the loop
  * If the node was idle at the start of the loop _and_ the `before_update` hook didn't cancel the update, do all the normal node I/O updates (delete/check/&c.)
* Loop over groups:
  * As with nodes, call the I/O layer `before_update` hook.  This function may also cancel the group update for this time through the loop
  * If all nodes in the group were idle at the start of the loop _and_ the `before_update` hook didn't cancel the update, do all the normal group I/O updates
* This is the end of the "regular I/O" update.
* If node or groups were idle at the end of the regular update, run low-priority "idle updates" on them.  This is going to be stuff like re-checking files on Nearline to see if the restored/released state saved in the DB is correct.
* Finally, run the I/O layer `after_update` hook on groups and then nodes.

**Housekeeping in the main loop:**
After all the I/O updates are complete for one iteration of the loop, several housekeeping tasks are performed each update loop:
* the worker pool scans for cleanly-exited workers (due to `peewee.OperationalError`) and restarts them
* If there are no workers in the pool, some time is spent executing I/O tasks in the main loop (see `serial_io()`).  This allows alpenhorn to fall-back to the non-multithreaded update system whenever it happens to have no workers to execute I/O tasks. (including when using a non-threadsafe DB, which will force the worker pool to be empty).
* The sleep at the bottom of the loop will now wake up early if a `global_abort` is triggered by a worker thread.

This rewrite of the main loop is essentially complete.  A few, relatively small, changes will happen in a future PR when the `auto_import` module is updated.  Additionally, most of the I/O update code in `update.py` will be moved into the I/O classes.

## Broken by this PR
This PR comments out all the actual I/O updates in the main loop so that I can run tests on the framework.  This will be resolved in the later DefaultIO PRs (#150 et seqq.)

## An afterthought
I think this is sufficient to declare this PR closes #60

### DIFF
--- a/alpenhorn/config.py
+++ b/alpenhorn/config.py
@@ -19,7 +19,8 @@ Example config:
     base:
         hostname: alpenhost
 
-    # Configure the data base connection with a peewee db_url
+    # Configure the database connection with a peewee db_url.  If using a database
+    # extension, that may require different data in this section.
     database:
         url: peewee_url
 
@@ -42,11 +43,19 @@ Example config:
 
     # Configure the operation of the local service
     service:
+        # Default number of worker threads
+        num_workers: 4
+
         # Minimum time length (in seconds) between updates
         update_interval: 60
 
         # Timescale on which to poll the filesystem for new data to import
         auto_import_interval: 30
+
+        # Maximum time (in seconds) to run serial I/O per update loop (these
+        # are I/O run tasks in the main thread, in cases when there are no
+        # worker threads
+        serial_io_timeout: 900
 
     # Set any configuration for acquisition type extensions
     acq_types:
@@ -73,8 +82,13 @@ log = logging.getLogger(__name__)
 config = None
 
 _default_config = {
-    "service": {"update_interval": 60, "auto_import_interval": 30},
     "logging": {"level": "warning", "module_levels": {"alpenhorn": "info"}},
+    "service": {
+        "auto_import_interval": 30,
+        "num_workers": 0,
+        "serial_io_timeout": 900,
+        "update_interval": 60,
+    },
 }
 
 
@@ -112,7 +126,8 @@ def load_config():
         with open(absfile, "r") as fh:
             conf = yaml.safe_load(fh)
 
-        config = merge_dict_tree(config, conf)
+        if conf is not None:
+            config = merge_dict_tree(config, conf)
 
     if not any_exist:
         raise RuntimeError("No configuration files available.")
--- a/alpenhorn/io/base.py
+++ b/alpenhorn/io/base.py
@@ -1,0 +1,210 @@
+"""BaseIO classes.
+
+Provides the basic infrastructure for StorageNode and StorageGroup I/O.
+
+These are very low-level classes.  Any module implementing the I/O class for
+something even remotely resembling a POSIX filesystem may be better served
+by subclassing from DefaultIO instead of from here directly.
+"""
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+import logging
+
+if TYPE_CHECKING:
+    from ..queue import FairMultiFIFOQueue
+    from ..storage import StorageNode, StorageGroup
+    from ..update import UpdateableNode
+
+log = logging.getLogger(__name__)
+
+
+class BaseNodeIO:
+    """Base class for StorageNode I/O modules in alpenhorn.
+
+    Parameters
+    ----------
+    node : StorageNode
+        the node
+    queue : FairMultiFIFOQueue
+        the task queue
+    config : dict
+        the parsed `node.io_config`. If `node.io_config` is None,
+        this is an empty `dict`.
+    """
+
+    # SETUP
+
+    def __init__(
+        self, node: StorageNode, config: dict, queue: FairMultiFIFOQueue
+    ) -> None:
+        self.node = node
+        self._queue = queue
+        self.config = config
+
+    def update(self, node: StorageNode) -> None:
+        """Update the cached StorageNode instance.
+
+                Called once per update loop on pre-existing I/O instances to
+        replace their `self.node` with a new instance fetched from
+        the database.  The new `node` instance reflects changes made
+        to the database record outside of alpenhornd.
+
+        None of `node.id`, `node.name`, `node.io_class`, or `node.io_config`
+        are different than the old `self.node`'s values.  (Changes in these
+        attributes cause alpenhorn to re-create the I/O instance instead of
+        calling this method.)
+
+        If called, this method is called before the `before_update` hook,
+        but it is not called during the main loop iteration that creates
+        the I/O instance.
+
+        Parameters
+        ----------
+        node : StorageNode
+            the updated node instance read from the database
+        """
+        self.node = node
+
+    # HOOKS
+
+    def before_update(self, idle: bool) -> bool:
+        """Pre-update hook.
+
+        Called each update loop before node updates happen.
+
+        Parameters
+        ----------
+        idle : bool
+            If False, updates for this node are going to be skipped this
+            update loop.
+
+        Returns
+        -------
+        do_update : bool
+            Whether to proceed with the update or not (skip it).  If
+            False, the update will be skipped.
+        """
+        # By default, we do nothing and allow the update to continue
+        return True
+
+    def idle_update(self) -> None:
+        """Idle update hook.
+
+        Called after a regular update that wasn't skipped, but only if,
+        after the regular update, there were no tasks pending or in
+        progress for this node (i.e. `self.idle` is True).
+
+        This is the place to put low-priority tasks that should only happen
+        if no other I/O is happening on the node.
+        """
+        # By default do nothing.
+        pass
+
+    def after_update(self) -> None:
+        """Post-update hook.  Called at the end of the update loop.
+
+        This method is called once per update loop, after all other processing
+        has happened on the node.
+        """
+        # Do nothing
+        pass
+
+
+class BaseGroupIO:
+    """Base class for StorageGroup IO modules in alpenhorn.
+
+    Parameters
+    ----------
+    group : StorageGroup
+        The group
+    config : dict
+        The parsed `group.io_config`. If `group.io_config` is None,
+        this is an empty `dict`.
+    """
+
+    # SETUP
+
+    def __init__(self, group: StorageGroup, config: dict) -> None:
+        self.group = group
+        self.config = config
+
+    def set_nodes(self, nodes: list[UpdateableNode]) -> list[UpdateableNode]:
+        """Set the list of local active nodes in this group.
+
+        This method is called to communicate to the I/O instance the list
+        of locally-active nodes.
+
+        This method is called each main loop, after regular
+        node I/O has completed but before any group I/O updates commence.
+
+        If group I/O cannot proceed with the supplied list of nodes,
+        implementations should raise ValueError with a message which will
+        be written to the log.
+
+        Otherwise, it may choose to operate on any non-empty subset of
+        `nodes`.  In this case it should return the list of nodes which has
+        been selected, and record the list locally, if needed.
+
+        Parameters
+        ----------
+        nodes : list of UpdateableNodes
+            local active nodes in this group.  Will never be empty.
+
+        Returns
+        -------
+        selected_nodes : list of UpdateableNodes
+            `nodes` or a non-empty subset of `nodes` on which group I/O
+            will be performed.
+
+                Raises
+                ------
+        ValueError
+            `nodes` was not sufficient to permit group I/O to proceed.
+        """
+        raise NotImplementedError("method must be re-implemented in subclass.")
+
+    # HOOKS
+
+    def before_update(self, idle: bool) -> bool:
+        """Pre-update hook.
+
+        This method is called once per update loop, before any other
+        processing happens on this group.
+
+        Parameters
+        ----------
+        idle : boolean
+                If False, the group update is going to be skipped.
+
+        Returns
+        -------
+        do_update : bool
+            Whether to proceed with the update or not (skip it).  If
+            False, the update will be skipped.
+        """
+        # By default, we do nothing and allow the update to continue
+        return True
+
+    def idle_update(self) -> None:
+        """Idle update hook.
+
+        Called after a regular update that wasn't skipped, but only if after
+        the regular update, there were no tasks pending or in progress this
+        group (i.e. self.idle is True).
+
+        This is the place to put low-priority tasks that should only happen
+        if no other I/O is happening on the group.
+        """
+        # By default do nothing.
+        pass
+
+    def after_update(self) -> None:
+        """Post-update hook.
+
+        This method is called once per update loop, after all other processing
+        has happened on the group but before the node `after_update` hooks
+        are called.
+        """
+        # Do nothing
+        pass
--- a/alpenhorn/storage.py
+++ b/alpenhorn/storage.py
@@ -3,16 +3,15 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING
 
-if TYPE_CHECKING:
-    import pathlib
-    from .acquisition import ArchiveFile
-
-
 import peewee as pw
 from peewee import fn
 
 from .db import EnumField, base_model
 from . import util
+
+if TYPE_CHECKING:
+    import pathlib
+    from .acquisition import ArchiveFile
 
 
 class StorageGroup(base_model):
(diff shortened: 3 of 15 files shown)


alpenhorn/io/Default.py
```diff
@@ -0,0 +1,24 @@
+"""Alpenhorn Default I/O classes.
+
+The Alpenhorn Default I/O classes largely re-create the legacy I/O behaviour
+of previous versions of Alpenhorn.  These I/O classes are used with StorageNodes
+and StorageGroups which do not explicitly specify io_class.
```
ketiltrout, line 5:
```suggestion
of previous versions of Alpenhorn.

These I/O classes are used by StorageNodes and StorageGroups which do not
explicitly specify `io_class` (as well as being used explicitly when `io_class`
has the value "Default").
```

alpenhorn/update.py
```diff
@@ -36,86 +39,283 @@
...
-        # Iterate over nodes and perform each update (perform a new query
-        # each time in case we get a new node, e.g. transport disk)
-        for node in st.StorageNode.select().where(st.StorageNode.host == host):
-            update_node(node)
+        # Used to remember things about active groups on this host
+        groups = dict()
+
+        # Iterate over nodes first
+        #
+        # nodes and are re-queried every loop iteration so we can
```
ketiltrout, line 81:
```suggestion
        # nodes are re-queried every loop iteration so we can
```

alpenhorn/update.py
```diff
@@ -36,86 +39,291 @@
...
+    if idle and do_update:
+        log.info(f'Updating node "{node.name}".')
+
+        # Check the integrity of any questionable files (has_file=M)
+        # XXX Commented out until future PR.
+        # update_node_integrity(node)
 
-    # Delete any upwanted files to cleanup space
-    update_node_delete(node)
+        # Delete any upwanted files to cleanup space
```
ketiltrout, line 283:
```suggestion
        # Delete any unwanted files to cleanup space
```

alpenhorn/update.py
```diff
@@ -36,86 +39,291 @@
...
+        pool.check()
+
+        # If we have no workers, handle some queued I/O tasks
+        if len(pool) == 0:
+            serial_io(queue)
+
+        # Check the time spent so far
         loop_time = time.time() - loop_start
-        log.info("Main loop execution was %d sec.", loop_time)
+        log.info(f"Main loop execution was {loop_time} sec.")
```
ketiltrout, line 161:
```suggestion
        log.info(f"Main loop execution was {util.pretty_deltat(loop_time)}.")
```
(which was the whole point of introducing that function at this point.)

alpenhorn/storage.py
```diff
@@ -1,18 +1,75 @@
...
 
-from .db import EnumField, base_model
-from . import util
+
+@overload
+def _get_io_instance(obj: StorageGroup) -> BaseGroupIO:
+    ...
+
+
+def _get_io_instance(obj):
```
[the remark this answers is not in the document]
ljgray: I think using `overload` is good

jrs65 (changes requested):
This looks good, but I do have a few questions that I'll chat with you in person about.

alpenhorn/storage.py
```diff
@@ -1,18 +1,75 @@
...
+    ...
 
-from .db import EnumField, base_model
-from . import util
+
+@overload
+def _get_io_instance(obj: StorageGroup) -> BaseGroupIO:
+    ...
+
+
```
ljgray, line 28:
```suggestion
```

alpenhorn/storage.py
```diff
@@ -1,18 +1,75 @@
...
 
-import peewee as pw
-from peewee import fn
+@overload
+def _get_io_instance(obj: StorageNode) -> BaseNodeIO:
+    ...
 
-from .db import EnumField, base_model
-from . import util
+
```
ljgray, line 23:
```suggestion
```

alpenhorn/io/base.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Load config if present
+        if node.io_config is None:
+            self.config = dict()
+        else:
+            self.config = json.loads(node.io_config)
+
+    def set_queue(self, queue: FairMultiFIFOQueue) -> None:
+        """Set the queue used for asynchronous I/O tasks."""
+        self._queue = queue
```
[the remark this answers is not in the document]
ljgray: I was in the middle of reviewing this when you added this review. This was something I was thinking as well - at the very least, `_queue` should be set to _something_ in the constructor so that we don't end up with an odd AttributeError, or there should be a `queue` property which raises a more informative exception if `_queue` hasn't been set.

alpenhorn/update.py
```diff
@@ -36,86 +39,291 @@
...
-        # Iterate over nodes and perform each update (perform a new query
-        # each time in case we get a new node, e.g. transport disk)
-        for node in st.StorageNode.select().where(st.StorageNode.host == host):
-            update_node(node)
+        # Used to remember things about active groups and nodes on this host
+        groups = dict()
+        updated_nodes = list()
+
+        # Iterate over nodes first
+        #
```
ljgray, line 81:
```suggestion
```

alpenhorn/update.py
```diff
@@ -36,86 +39,291 @@
...
-    if node.suspect:
-        log.debug('Skipping suspected node "%s".', node.name)
+    This function is only called when alpenhorn has no worker threads.  It runs
+    I/O tasks in the main loop for a limited period of time.
+    """
+    start_time = time.time()
 
-    log.info('Updating node "%s".', node.name)
+    # Handle tasks for, say, 15 minutes at most
+    while time.time() - start_time < 900:
```
ljgray, line 187: Maybe this timeout should be configurable

alpenhorn/util.py
```diff
@@ -111,6 +111,48 @@ def get_hostname() -> str:
...
+    if seconds < 0:
+        raise ValueError("negative time delta")
+
+    hours, seconds = divmod(seconds, 3600)
+    minutes, seconds = divmod(seconds, 60)
+
+    if hours > 0:
+        return f"{int(hours)}h {int(minutes):02}m {int(seconds):02}s"
+    if minutes > 0:
+        return f"{int(minutes)}m {int(seconds):02}s"
```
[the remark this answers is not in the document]
ketiltrout: Sure, I think that would work.

alpenhorn/io/base.py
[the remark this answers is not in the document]
ketiltrout: There's only one queue for the daemon as a whole (It's the task scheduler).  A new one can't be created for each StorageNode.

I think this question might be partly due to not having enough context for the full I/O framework.

The queue is never re-assigned, there's only ever one queue, created in the alpenhornd start-up.  Would it be better to assign the queue instance to a global and access it that way?

This part of the system doesn't seem very nice, and I'm not opposed to changing it somehow.

We could merge this call into the initialiser, but it wouldn't save any calls.  Instead of doing:
```
# Set the queue (I/O instance is automatically created the first time `node.io` is accessed)
node.io.set_queue(queue)
# Run an integrity check
node.io.check(filecopy)
```
it would be:
```
# Create the I/O instance
node.create_io(queue)
# Run an integrity check
node.io.check(filecopy)
```
_Most_ methods in the I/O framework don't use the queue, and don't need `set_queue` to be called, but I think in practice I think it's rare for I/O methods to be called without that happening.

alpenhorn/update.py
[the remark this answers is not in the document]
ketiltrout: Yes.

alpenhorn/storage.py
ketiltrout: These lines are added by `black`.  We can't remove them.

ljgray: Oh, interesting. My bad

alpenhorn/storage.py
```diff
@@ -126,6 +235,15 @@ class StorageNode(base_model):
     avail_gb = pw.FloatField(null=True)
     avail_gb_last_checked = pw.DateTimeField(null=True)
     notes = pw.TextField(null=True)
+    io_config = pw.TextField(null=True)
```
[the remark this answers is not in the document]
ketiltrout: Yeah, I've though about that.

For now, I think it's best to keep this as a simple TextField.  The JSONField is most helpful when writing to the field. My plan is to revisit this when I get around to updating the client.

alpenhorn/config.py
```diff
@@ -19,7 +19,8 @@
     base:
         hostname: alpenhost
 
-    # Configure the data base connection with a peewee db_url
+    # Configure the database connection with a peewee db_url.  If using a database
```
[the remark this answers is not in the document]
ketiltrout: With monospaced fonts, two spaces after a full stop aid in legibility.

alpenhorn/update.py
ketiltrout: I prefer having the blank line.  It acts as a paragraph break for the comment.  (Not that this is a terribly long comment, though.)

alpenhorn/io/base.py
```diff
@@ -0,0 +1,173 @@
...
+    # HOOKS
+
+    def before_update(self, idle: bool) -> bool:
+        """Pre-update hook.
+
+        Called each update loop before node updates happen.
+
+        Parameters
+        ----------
+        idle : bool
```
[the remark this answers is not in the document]
ketiltrout: That seems like an argument for not adding them if they don't already exist, but is there a reason to _remove_ them?

alpenhorn/storage.py
```diff
@@ -22,12 +79,42 @@ class StorageGroup(base_model):
...
     notes : string
         Any notes about this storage group.
+    io_config : string
+        An optional JSON blob of configuration data interpreted by the
+        I/O class
+
+    If `io_class` includes at least one `.`, then it's assumed to be
+    the full import path to a class.
+
+    Otherwise, the I/O module is assumed to be part of the
```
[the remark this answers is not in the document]
ketiltrout: The DB entries can't be lower cased unless you want all the classes to have lowercase names as well.  Is that what you're suggesting?

ketiltrout:
I've made a fairly major update here to address a number of comments on this PR:

I've created new container objects (in `update.py`) called `UpdateableNode` and `UpdateableGroup`.  They contain both the `StorageNode/Storagegroup` and the I/O instance and removes the circular references between the two.

These new objects take infrastructural elements that I had sprinked around the code base for lack of better places to put them.  Some of these has been put in the I/O base classes (in `io/base.py`) as base I/O class methods which were not meant to be re-implemented in subclasses, which was awkward.

The parts now unified in these new classes are:
* The I/O class loading and instantiation (`_get_io_class`) which I had put in `storage.py`
* The I/O config parsing from `io/base.py`
* The node/group idle checking from `io/base.py`

It also provides a better way to do I/O instantiation (including removing the awkward `set_queue` method, and node-vetting functionality of the group I/O classes.

These classes also now subsume the `update_node`/`update_group` functions (e.g. `update_node()` is now `UpdateableNode.update()`) and their sub-functions, though most of those will be moved into the new classes as they're updated.

The UpdateableNode/Group instances persist through main loop iterations, which also means the I/O classes do to, which I think is a more expected behaviour.  The top-level containers have logic to re-initialise the I/O classes if necessary.

alpenhorn/storage.py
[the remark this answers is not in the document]
ketiltrout: To make the DB entries lower case would also require lower-casing the class names (`DefaultNodeIO` &c.)  Is that what you're suggesting?

alpenhorn/util.py
```diff
@@ -111,6 +111,48 @@ def get_hostname() -> str:
...
+    """
+
+    # Reject weird stuff
+    try:
+        seconds = float(seconds)
+    except (TypeError, ValueError):
+        raise TypeError("non-numeric time delta")
+
+    if seconds < 0:
+        raise ValueError("negative time delta")
```
ketiltrout, line 142: Is this bad?

It's possible alpenhornd could end up having a negative time delta due to some sort of clock skew on the host.  In that case, would it be better to simply print the nonsensical delta and leave it at that, rather than crashing?

ketiltrout: Yeah, I think I'll remove it.

alpenhorn/update.py
```diff
@@ -36,104 +45,603 @@
...
-def update_node_active(node):
-    """Check if a node is actually active in the filesystem"""
+        # Within the module, find the class
+        try:
+            class_ = getattr(module, io_name)
+        except AttributeError as e:
+            raise ImportError(
+                f'I/O class "{io_name}" not found in module "{io_module}". '
+                f"Required for {obj_type} {self.name}."
+            ) from e
```
jrs65, line 165: I would be pretty tempted to move this class resolution code into alpenhorn startup, i.e. build up a dict of enabled io class name -> implementation from the default alpenhorn ones and those provided by plugins.

I think having this in here pollutes the logic a bit, and it means that you could be potentially running imports within the main loop of alpenhorn which could cause issues. It feels like it would be much better to do this upfront when the service starts along with all the other config/plugin type stuff.

jrs65 (changes requested):
I think this is much improved. Those changes have not only clarified the ownership issues but have really made the structure of what it's doing much clearer.

I had one suggestion, but you should probably feel free to ignore it.

alpenhorn/update.py
ketiltrout: I get what you're saying.

I think effectively, in the normal case where available nodes don't change over the course of an `alpenhornd` run, this already happens for nodes, because all node I/O instantiation happens at the top of the first update loop, before the node update loop starts, and then doesn't change after that.

This is different than the Info class modules, which would be loaded at a surprising time in the middle of an import task if they weren't preloaded on start-up (and so it makes a lot of sense to do the preload). So, I think it's less of an issue here for I/O classes.

I don't think that just blindly preloading all I/O classes is a good idea, because it would require installations to have all their weird third-party I/O modules installed, even on hosts where nodes using those weird I/O modules don't exist.

As a half-way solution, why don't I also move the group-I/O instantiation to the top of the loop (currently happens after node updates are done).  Then all the module loading is done early, at the start of the first update loop, except for cases where new nodes/groups show up later with hitherto unseen I/O classes.

jrs65: Yeah, your half-way solution seems reasonable, and this isn't the kind of structuring that is embedded so deeply it can't be changed later.

To clarify though, I wasn't expecting that you would pre-load or require all possible IO classes, just that you would specify them through the plugin system, and thus the local config would only need to specify (and thus load) IO classes for all the types of nodes it expects to see.

jrs65 (approved):

PR #149, opened 2023-02-23
Rewrite 6/14: The "import-detect" Extension and removal of Type and Info classes

ketiltrout:
This PR is concerned with implementing the "file import detection" framework for the daemon.  I think this is the last of the "structural" PRs.  Subsequent PRs in this rewrite will deal primarily with changes to I/O code. 

## Motivation
The ultimate goal of this PR is to produce the infrastructure needed by the CHIME alpenhorn extensions over in [alpenhorn-chime](https://github.com/chime-experiment/alpenhorn-chime/).

This PR does the following:
* removes `AcqType`, `FileType` and all the Info class framework (all moved to `alpenhorn-chime`).
* replaces the type/info-based detection system with a simple function call plus callback scheme provided by third-party extensions (like `alpenhorn-chime`)
* moves the old example type extension `alpenhorn/generic.py` to `examples/pattern_importer.py`, and updates it to work with this new system

## Removal of the Info framework
I've removed all reference to info classes from alpenhorn.  They were an integral part of alpenhorn-1, but in alpenhorn-2 they served two purposes:
 * they had a `detect` function which was passed an acq or file name and returned True or False to indicate a file which needed to be imported by alpenhorn
 * because they were peewee models, when importing a file, alpenhorn would also call a `new` class method to generate a new record in these tables.

The first of these features has been replaced by a new "import-detect" extension type which provides a simple function which will perform the detection step of the import.  See the "The 'import-detect' Extension" section below.

The second of these functions is replaced with an optional post-import hook, which removes the awkwardness of requiring alpenhorn to add rows to tables it knows nothing about.  See the "The post-import Callback" section below.

## Removal of `AcqType` and `FileType`
While CHIME makes heavy use of `AcqType` and `FileType` to manage our data, in alpenhorn their use was solely to determine which Info tables were available to perform import detection.  With the removal of info classes, they no longer have a use in alpenhorn.  I've moved them to `alpenhorn-chime` where they've been re-implemented (like the `ArchiveInst` table was).

Removal of these two tables also means the `acq_types` and `file_types` extensions are no longer needed, and they have been removed from `extensions.py`, as well as the `register_type_extensions` call that was being made in `service.py`.

## The "import-detect" Extension

In place of all the above is a new extension type called "import-detect".  Each "import-detect" extension returns (via `register_extension`) a single callable object, which is the "detect" function used during file import.

(This is exactly what `alpenhorn-chime` is: an alpenhhorn import-detect extension.)

The detect function is passed a `pathlib.Path` pointing to the file to import and the `UpdateableNode` containing the candidate data file.  The function must determine if the path points to a file that alpenhorn should import.  It must return a 2-tuple:
 * The first element is either `None` to indicate detection failed (i.e. the path does not point to a data file needing import), or else it's the acquisition name, which must be a portion of the path passed in (with the file name becoming the remaining portion of the path).
 * The second element is an optional callback function which will be called after the file has been imported (i.e. after `ArchiveAcq` `ArchiveFile`, `ArchiveFileCopy` records have been generated for the file.  If no post-import callback is needed, this may be `None`.  (In implementing this new system, I've discovered `functools.partial` to be a useful thing to retrun to alpenhorn as a callback because it allows the detect function to pass data to a callback.)  See the following section for more details

Multiple "import-detect" extensions may be loaded.  In that case, the import code tries each in extension order until one of them reports a successful match.

alpenhorn _will_ run without any "import-detect" extensions loaded, but will be unable to import files in that case.  (Attempts to import files will result in an error message).

## The Post-Import Callback
When provided, alpenhorn will pass to the callback the following parameters:
* the `ArchiveFileCopy` of the newly imported file
* if this import created it, the `ArchiveFile` for the file (or `None` if it already existed)
* if this import created it, the `ArchiveAcq` for the file (or `None` if it already existed)
* the `UpdateableNode` on which this import took place.

The `ArchiveFile` and `ArchiveAcq` of any imported file may be obtained via the `ArchiveFileCopy`; newly-created instances are passed to the callback so the callback knows when they are new or not.  These two parameters could be replaced by booleans without loss of information, but I think it's more direct to do it like this.

The value returned from the callback is ignored.

## The "pattern-importer" example extension

The regex/glob-based example extension formerly found in `alpenhorn/generic.py` (which was an example of an "acq_types"/"file_types" extension) has been moved to `examples/pattern-importer.py` and updated to be an "import-detect". 

It's not used _yet_ but this example extension will also eventually be used in the end-to-end test in `tests/test_service.py`.

## Changes to `auto_import`
The changes here are somewhat performative: they _are_ the code changes needed to use the new info class system, but the `auto_import` code doesn't really work yet within the new framework.  A subsequent PR in this series will transition the import  code to use the new task queue.  As part of that, this code will get fixed.  Despite that, it's good to make this change here to show how the changes to info classes affect the calling code.


alpenhorn/auto_import.py
```diff
@@ -225,14 +312,32 @@ def add_acq(acq_type, name, node, comment=""):
     if ac.ArchiveAcq.select(ac.ArchiveAcq.id).where(ac.ArchiveAcq.name == name).count():
         raise AlreadyExists('Acquisition "%s" already exists in DB.' % name)
 
+    # Extended acq class
+    ExtendedArchiveAcq = extensions.extended_table(ac.ArchiveAcq)
+
+    # Get the extended table data, if any
+    ext_data = _get_extended_table_data(
+        ExtendedArchiveAcq, acq_name, acq_type, file_name, None, node, path
```
ketiltrout, line 320: The `None` here should be the `FileType`, but that doesn't exist at this point because this call is happening too early in this version of the code.

A later PR will fix this.

tests/test_info_base.py
```diff
@@ -0,0 +1,268 @@
...
+    assert not info_base.has_model()
+    assert not _NoInfo.has_model()
+    assert acq_info_base.has_model()
+    assert file_info_base.has_model()
+    assert GenericAcqInfo.has_model()
+    assert GenericFileInfo.has_model()
+
+
+def test_is_acq():
+    """Test is_acq()"""
```
ljgray, line 30: Assuming this will be implemented in future PRs

ljgray (dismissed):
Looks good overall. A couple of comments about style, mainly about CapWords vs snake_case for class names.

The implementation here looks good overall to me, but I'm not familiar with the previous implementation in alpenhorn 1 so I'm generally looking at it as a completely new thing and not so much as changes to the previous verison, which I figure is generally what you're going for here anyway.

alpenhorn/acquisition.py
```diff
@@ -1,157 +1,265 @@
...
-from .config import ConfigClass
+from . import config
 from .db import base_model
 
 log = logging.getLogger(__name__)
 
 
-class AcqType(base_model):
-    """The type of data that is being taken in the acquisition.
+class type_base(base_model):
```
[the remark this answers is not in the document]
ketiltrout: Taking inspiration from things like `db.base_modeI`, I've generally used snake case for the names of abstract base classes which are meant to never be instantiated (like this), to better distinguish them from concrete classes.

Is that frowned upon in python?

tests/test_info_base.py
ketiltrout: That's odd...

alpenhorn/acquisition.py
```diff
@@ -1,157 +1,265 @@
...
+            for name in path.parents:
+                if str(name) != ".":
+                    if info_class.is_type(name, node):
+                        return type_, str(name)
+
+        # No match
+        return None, None
+
+    def info(self):
+        """Return the Info class for this AcqType.
```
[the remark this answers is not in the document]
ketiltrout: That could work with `AcqType` but `FileType.info()` needs an argument (the AcqType).

alpenhorn/extensions.py
```diff
@@ -119,48 +218,54 @@ def database_extension() -> dict:
...
+    name = table.__name__
+    if name not in _model_ext:
+        raise TypeError(f"{table} is not an extensible table model")
 
-        acq_type.set_config(conf)
+    # Return base table if not extended
+    if len(_model_ext[name]["fields"]) == 0:
+        return table
 
+    class extended_table(table):
```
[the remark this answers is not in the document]
ketiltrout: This one should probably be changed.  (This is a concrete class.)

tests/test_info_base.py
ketiltrout: OK.  I don't know what happened to it.  Presumably a casualty of one of the infinite rebases. I've re-implemented the test.

alpenhorn/acquisition.py
```diff
@@ -1,157 +1,263 @@
...
+        # Get the import_error action for this type
+        self._info_import_errors = config.info_import_errors(
+            self.name, is_acq=isinstance(self, AcqType)
+        )
 
-        if len(missing_acq_types):
-            raise RuntimeError(
-                "AcqTypes %s have no registered handler." % repr(missing_acq_types)
-            )
+    def _get_info_class(self, acqtype: AcqType | None, is_acq: bool) -> info_base:
```
ketiltrout, line 80: This function and the corresponding `_get_io_class` from the previous PR do essentially the same thing:
* Fetch the value of `info_class` (or `io_class`) from the database record
* Import the module that it specifies
* Return the named class from the imported module

This works fine as-is, and gives alpenhorn broad flexibility to deal with new modules/classes being introduced to DB records after start-up.

However, I've though up another possibility that I think would do okay:
* restrict the `info_class` field to contain only a class name (or, in the case of `io_class`, the "base name" for the I/O class, like "Nearline" or "Default").
* Require third party module which implement an Info Class or I/O class to appear in the extensions list
* The `register_extension` function of one of these classes needs to return a list of InfoClasses and/or I/O modules which they implement.
* InfoClasses declared by `register_extension` get added to a container module `alpenhorn.info` and the class names found in the `info_class` for Acq/FileTypes then refer to classes in that namespace.
* Similarly I/O modules are added to `alpenhorn.io`.

This has the benefit that all module import happens at start-up.  And it also removes the possibility for arbitrary code execution via database injection attack.  However, something would need to be figured out with how to deal with classes which are declared in the DB but not loaded from an extension.

ketiltrout (commented):
I have though of two questions I'd like advice on.

ketiltrout:
This is a major rewrite of this PR, which I think makes things better by making alpenhorn simpler, though paring away weird CHIME-isms which don't contribute to alpenhorn's goals.

Instead of the complex Info-table framework and the model-extension stuff I had written, now I've just dropped `Info` tables completely from alpenhorn because, like `ArchiveInst`, alpenhorn doesn't care about them and shouldn't be the one dealing with them.

And then, by dropping info tables, we can also drop `AcqType` and `FileType`, because all they were doing for alpenhorn was telling it which info classes to interrogate for which imported files, so without info tables, they have nothing to do.  It's better to let the third parties, decide how to organize their data rather than arbitrarily forcing something on them for which we have no need.

(For CHIME, all of these dropped things are moved/reimplemented in `alpenhorn-chime`).

Overall I think it's cleaner, more intuitive, and more flexible without losing anything CHIME needs.

ljgray (commented):
This looks good to me

ljgray (approved):

PR #150, opened 2023-03-01
Rewrite 7/14: DefaultIO housekeeping + integrity check

ketiltrout:
_This is the first of four PRs to move I/O operations to the `DefaultIO` framework._

## Overview

Three parts of the update loop are handled here:
* node check (checking for the `ALPENHORN_NODE` file).
* free space update
* integrity check of file copies with `has_file=='M'`

Also, I've made a change to the way third-party I/O classes are loaded; see the following section.

## I/O module loading

Following discussion with Richard, this PR modifies how third-party I/O modules are found by alpenhorn from the behaviour originally introduced in Rewrite part 5/14 (#148).

The old behaviour was to simply have the full import path given in, say, `StorageNode.io_class`.  Instead of doing that, this PR introduces a new extension type called "io-module" which extensions should use to provide alpenhorn with third-party I/O modules.   The benefit here is all I/O modules are imported at start-up, rather than the first time alpenhorn sees a particular storage object.

The value associated with the "io-modules" key returned by `register_extensions` is a dictionary whose keys are I/O module names and whose values are the modules themselves.  The third-party modules may not have names that duplicate any of the internal modules in `alpenhorn.io`, nor may multiple extensions provide modules with the same name.

The module name must follow the convention that the internal modules use: if there is a node or group with `io_class` equal to, say, `MyCustomIO`, then there must be an I/O module with name (=dict key) `mycustomio` (i.e. lowercased-version) and that I/O module must define the class `MyCustomIONodeIO` or `MyCustomIOGroupIO` as appropriate.

The prior behaviour of allowing a full import path in the `io_class` column is no longer allowed.

A missing I/O class at runtime is not fatal: if that happens, `alpenhornd` will report an error in the log and act as if the node or group that requested the I/O class is not active (i.e. the affected node/group is ignored during the update loop).  Because an attempt at I/O class instantiation happens every time through the loop, the error message about the missing class will be repeated each time through the loop.  This is intentional, to make it clear to operators why the node/group is not being updated.

## Node check

This is a relatively minor change.  The function `util.alpenhorn_node_check` becomes the I/O method `node.io.check_avail` (i.e. `DefaultNodeIO.check_avail`) but is essentially unchanged and is called in the same place in `update.update_node_active`.

## Free space

The outer function, `update.update_node_free_space` is moved to `util.update_avail_gb` because it will be called from other places beside the main loop in the future.  The function is changed to call `node.io.bytes_avail` where the actual stat occurs to fetch the amount of free space.

This PR also adds a boolean parameter `fast` to `update_avail_gb` which is passed on to `node.io.bytes_avail`.  The invocation of `update_avail_gb` from the main loop has `fast=False`, but the other invocations will have `fast=True`.  This boolean will allow I/O modules for which finding the free space is slow to skip the optional `fast=True` invocations.

The code also permits `node.io.bytes_avail` to return `None`, meaning the free space couldn't be determined (or doesn't make sense).

## Integrity check

Running the integrity check is the first asynchronous task added to this rewrite (the function `check_async`).  The asynchronous part of the check is put in `alpenhorn/io/_default_asyncs.py`, mostly for tidiness and to keep the size of `Default.py` manageable.

The integrity check is as follows:
* Does the file exist?  If not set `has_file='N'` and we're done.
* Next, is the file size correct?  If not set `has_file='X'` and we're done.  This is a new step for this PR which avoids having to calculate the MD5 hash in some instances.
* Calculate the MD5 and check it against the DB to determine `has_file='Y'` or `has_file='X'`

## Other changes
* The unused function `util.is_md5_hash` is deleted.
* The parameter `cmd_line` is removed from `util.md5sum_file`.  It was never set to anything other than the default.  Removing it removes an unnecessary dependency on an external program.

### DIFF
--- a/alpenhorn/extensions.py
+++ b/alpenhorn/extensions.py
@@ -4,7 +4,7 @@ Extensions are simply python packages or modules providing extra functionality
 to alpenhorn. They should be specified in the `'extension'` section of the
 alpenhorn configuration as fully qualified python name. They must have a
 `'register_extension'` function that returns a `dict` specifying the extra
-functionality they provide. There are currently two supported keys:
+functionality they provide. There are currently three supported keys:
 
 `database`
     A dict providing capabilities of a database extension.  See the db module
@@ -44,6 +44,24 @@ functionality they provide. There are currently two supported keys:
 
     If multiple `import-detect` extensions are provided, they will be called in the
     order given in the config file until one of them indicates a successful match.
+`io-modules`
+    A dict providing I/O modules to augment the default modules provided in
+    `alpenhorn.io`.  Each I/O module should have a node I/O or group I/O class
+    (or both).  I/O class names and dict keys must adhere to the following naming
+    conventions:
+      * For a StorageNode with `io_class` equal to "IOClassName", the node I/O class
+        implementing this I/O class must be called `IOClassNameNodeIO`.
+      * Similarly, a StorageGroup with `io_class == "IOClassName" must be named
+        `IOClassNameGroupIO`.
+      * In the `io-modules` dict, the key whose value is a module containing either
+        (or both) of the above classes must be "ioclassname" (i.e. equivalent to the
+        `io_class` of the group and/or node after conversion to lower case).  So,
+        the above example classes would both be in a module associated with the dict
+        key `ioclassname`.
+
+    Multiple `io-modules` extensions may be provided; no two extensions may provide
+    the same dict keys, nor may any extension provide a key which is the name of an
+    existing `alpenhorn.io` submodule.
 
 If other keys are present in the dictionary returned by `register_extension`, they
 are ignored.
@@ -58,6 +76,7 @@ from . import config
 
 if TYPE_CHECKING:
     import pathlib
+    from types import ModuleType
     from collections.abc import Callable
 
     from .alpenhorn import ArchiveAcq, ArchiveFile
@@ -77,6 +96,7 @@ log = logging.getLogger(__name__)
 # Internal variables for holding the extension references
 _db_ext = None
 _id_ext = None
+_io_ext = dict()
 
 
 def load_extensions() -> None:
@@ -97,9 +117,7 @@ def load_extensions() -> None:
     TypeError
         `register_extension` provided data of the wrong type.
     ValueError
-        More than one database extension was specified in the alpenhorn
-        config or the object provided by an import-detect extension was
-        not callable.
+        The data returned by register_extension was not usable.
     """
 
     # Initialise globals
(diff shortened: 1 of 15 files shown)


alpenhorn/io/Default.py
```diff
@@ -7,17 +7,144 @@
...
 
 log = logging.getLogger(__name__)
 
 
 class DefaultNodeIO(BaseNodeIO):
     """A simple StorageNode backed by a regular POSIX filesystem."""
 
+    # I/O METHODS
+
+    def bytes_avail(self, fast: bool = False) -> int:
```
ketiltrout, line 35:
```suggestion
    def bytes_avail(self, fast: bool = False) -> int | None:
```

alpenhorn/io/base.py
```diff
@@ -95,6 +97,85 @@ def after_update(self) -> None:
         # Do nothing
         pass
 
+    # I/O METHODS
+
+    def bytes_avail(self, fast: bool = False) -> int:
```
ketiltrout, line 102:
```suggestion
    def bytes_avail(self, fast: bool = False) -> int | None:
```

alpenhorn/io/Default.py
```diff
@@ -7,17 +7,144 @@
...
+
+    def bytes_avail(self, fast: bool = False) -> int | None:
+        """bytes_avail: Return amount of free space (in bytes) of the node, or
+        None if that cannot be determined.
+
+        Does not account for space reserved via reserve_bytes().
+
+        Parameters
+        ----------
+        fast : bool
```
ljgray, line 43:
```suggestion
        fast
```
I think the convention when using type annotations is to not include the type in the docstring. This would apply everywhere

alpenhorn/util.py
```diff
@@ -111,6 +98,46 @@ def get_hostname() -> str:
...
+    if num < 2**10:
+        return f"{num} B"
+
+    for x, p in enumerate("kMGTPE"):
+        if num < 2 ** ((2 + x) * 10):
+            num /= 2 ** ((1 + x) * 10)
+            return f"{num:.1f} {p}iB"
+
+    # overflow
+    raise OverflowError("a suffusion of yellow")
```
ljgray, line 138: I like it but the exception should _probably_ be a bit more descriptive...

ljgray (changes requested):

alpenhorn/util.py
ketiltrout: Is it even worth having an exception here?  This function is only ever used for logging.  Do we want alpenhorn crashing when it discovers it's logging nonsense?

ljgray: We probably want something to happen if it's logging garbage so that someone notices it sooner rather than later, like an alert of some sort, but I'm not really sure how we'd implement that. If it is logging garbage it might be indicative of a larger issue, so if it eventually crashes from something else it would be rough to go back and find out that we can't use the logs to figure out what happened...

ketiltrout: In the end I just removed this exception.  I think it's best just to go on with logging in this case and let whatever caused the unreasonably large number do the crashing, if that needs to happen.

ketiltrout:
FYI: Per the updated description, I've added to this PR a change in the way I/O loading works.  (There's now an "io-module" extension that provides third-party I/O modules.)

ljgray (approved):
This looks pretty straightforward, one small change but approved otherwise

alpenhorn/extensions.py
```diff
@@ -199,3 +243,38 @@ def import_detection() -> list[ImportDetect]:
...
+    # Loading the I/O base is not allowed
+    if name == "base":
+        return None
+
+    # Try to load the module from alpenhorn.io
+    try:
+        return importlib.import_module("alpenhorn.io." + name)
+    except ImportError:
+        # No alpenhorn.io module, maybe it's an extension module
+        return _io_ext.get(name)
```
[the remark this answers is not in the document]
ketiltrout: `dict.get` never raises `KeyError`. The second parameter defaults to `None` if not specified.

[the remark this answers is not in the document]
ljgray: Oof, that's embarrassing. My bad

PR #151, opened 2023-03-01
Rewrite 8/14: DefaultIO file deletion

ketiltrout:
This is a relatively straightforward PR which mostly moves the guts of `update.update_node_delete` into `alpenhorn.io._default_asyncs.delete_async` which gets shoved into a task by `node.io.delete`.

I can't tell if it is for sure, but it seems weird to make an asynchronous task to delete a single file, so I've made the remaining code in `update.update_node_delete` group a bunch of them together and pass that to the I/O layer to delete together in the same task.

The bunch size I've chosen is 10.  Don't really know if that's a good number.  It's in the ballpark of the number of files that alpenhorn-1 tends to delete at a time (which, in turn, is related to the number of files it can transfer per hour).

I'm willing to entertain different values for the bunch size, including the value 1 (i.e. get rid of the bunching), but it seems like 1 or 10 or 100 are sort of the kind of numbers we should be considering here.

### DIFF
--- a/alpenhorn/io/_default_asyncs.py
+++ b/alpenhorn/io/_default_asyncs.py
@@ -2,11 +2,13 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING
 
+import errno
 import logging
+
+from ..archive import ArchiveFileCopy
 
 if TYPE_CHECKING:
     from .base import BaseNodeIO
-    from ..archive import ArchiveFileCopy
     from ..task import Task
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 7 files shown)


ljgray (approved):
I'm not sure how much the grouping is going to matter, it's really just a trade off of fewer longer-running tasks vs. more shorter-running tasks. I guess for really large operations having more threads might be marginally faster, since each individual thread deletes its files sequentially? I agree that we probably don't want to be starting new tasks for every single file that needs to be deleted, so 10 seems like an OK number to me (based on not a whole lot)

PR #152, opened 2023-03-01
Rewrite 9/14: I/O util (pull request inner functions)

ketiltrout:
This PR primarily moves parts of the file transfer code out of `update.py` and into `alpenhorn/io/ioutil.py`.  My original thought with these bits was that they would be useful to other I/O types other than the `Default` I/O, but that hasn't actually happened in this rewrite.  Nevertheless, it's good to split up the unmanageably long `update_node_requests`.

## Updates to previous rewrite PRs
This PR removes my ham-fisted attempt to create an auto-updating `CurrentTimestampField` (using MySQL's `ON UPDATE CURRENT_TIMESTAMP` clause) from way back in part 4 (#147) which wasn't going to work for a number of reasons.  In it's place is a simple `DateTimeField` which gets explicitly updated when necessary.

The only potential downside here is that changes made to the `ArchiveFileCopy` table outside of alpenhorn won't benefit from automatic updates of this cell, but I don't think we're ever updating that table outside of alpenhorn anyways.

## File transfer methods

There are three functions handling the various transfer methods moved into `ioutil`:
* bbcp
* rsync
* hardlink

These three functions all return an "ioresult" dict with standard-ish keys indicating the result of the transfer.

## Post transfer DB update
The final functionality moved to `ioutil` is a function that essentially takes the `ioresult` provided by the three transfer functions, determines whether the transfer succeeded or failed, and then and updates the database to reflect what happened. (This is `copy_request_done`.)

## Updates to `update.py`
Even though it's still commented out, I've updated `update.update_node_requests` to use these new functions.  In the next PR, `update_node_requests` will be converted to the new I/O system and these changes will be incorporated there.

## Changes to `util.py`
* `run_command` now takes an optional `timeout` parameter and will kill the subprocess if a specified timeout is exceeded.  I don't know why, but ever since I moved alpenhorn to the `chimedat` user, we've been running into issues where the `bbcp` transfer rate drops to zero without the transfer failing.  Not sure what's going on there, but this will work around the problem nicely.
* `command_available` has been deleted.  This is just a worse version of `shutil.which` (now used in its place).  Furthermore, `distutils`, which this function was based on is deprecated and scheduled for removal in 3.12.

### DIFF
--- a/alpenhorn/util.py
+++ b/alpenhorn/util.py
@@ -10,49 +10,52 @@ from . import config
 log = logging.getLogger(__name__)
 
 
-def run_command(cmd, **kwargs):
+def run_command(
+    cmd: list[str], timeout: float | None = None, **kwargs
+) -> tuple(int | None, str, str):
     """Run a command.
 
     Parameters
     ----------
-    cmd : array
+    cmd : list of strings
         A command as a list of strings including all arguments.
-    kwargs : dict
-        Passed directly onto `subprocess.Popen.`
+    timeout : float or None
+        Number of seconds to wait before forceably killing the process,
+        or None to wait forever.
+
+    Other keyword args are passed directly on to subprocess.Popen
 
     Returns
     -------
-    retval : int
-        Return code.
-    stdout_val : string
+    retval : int or None
+        Return code, or None if the process was killed after timing out.
+    stdout : string
         Value of stdout.
-    stderr_val : string
+    stderr : string
         Value of stderr.
     """
 
     import subprocess
 
-    log.debug('Running command "%s"', " ".join(cmd))
+    log.debug(f"Running command [timeout={timeout}]: " + " ".join(cmd))
 
     # run using Popen
     proc = subprocess.Popen(
         cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs
     )
-    stdout_val, stderr_val = proc.communicate()
-    retval = proc.returncode
+    try:
+        stdout_val, stderr_val = proc.communicate(timeout=timeout)
+        retval = proc.returncode
+    except subprocess.TimeoutExpired:
+        proc.kill()
+        stdout_val, stderr_val = proc.communicate()
+        retval = None
 
     return (
         retval,
         stdout_val.decode(errors="replace"),
         stderr_val.decode(errors="replace"),
     )
-
-
-def command_available(cmd):
-    """Is this command available on the system."""
-    from distutils import spawn
-
-    return spawn.find_executable(cmd) is not None
 
 
 def md5sum_file(filename: str, hr: bool = True) -> str:
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,12 @@ def pytest_configure(config):
 
     config.addinivalue_line(
         "markers",
+        "run_command_result(ret, stdout, stderr): "
+        "used on tests which mock alpenhorn.util.run_command to "
+        "set the desired return value for the mocked call.",
+    )
+    config.addinivalue_line(
+        "markers",
         "alpenhorn_config(*config_dict): "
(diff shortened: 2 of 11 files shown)


alpenhorn/util.py
```diff
@@ -15,36 +15,49 @@
 log = logging.getLogger(__name__)
 
 
-def run_command(cmd, **kwargs):
+def run_command(cmd, timeout=None, **kwargs):
```
ketiltrout, line 18: Needs type hinting

ljgray:
> There are two constants defined in this file, `PULL_TIMEOUT_BASE` and `PULL_BYTES_PER_SECOND` which determine the length of time before the transfer is killed. I've indicated that these could be put in the alpenhorn config, but I'm not sure it's worth it. I'm looking for opinions.

Personally I lean towards having them in config, because if we ever end up wanting to change them for some reason it's more of a pain to hunt down where they're defined. Also, depending on how the deployment is done, a change would just have to involve redeploying the config (assuming it's an ansible play)

alpenhorn/util.py
```diff
@@ -15,36 +15,51 @@
...
-        Value of stdout.
-    stderr_val : string
-        Value of stderr.
+    Returns a three-element tuple containing:
+        retval : int or None
+            Return code, or None if the process was killed after timing out.
+        stdout : string
+            Value of stdout.
+        stderr : string
+            Value of stderr.
```
ljgray, line 41:
```suggestion
    retval : int or None
        Return code, or None if the process was killed after timing out.
    stdout : string
        Value of stdout.
    stderr : string
        Value of stderr.
```
Since python packages multiple returns as a tuple by default, I think we can just list all the returns without having to specify that they're in a tuple (especially now with the extra type hinting)

alpenhorn/util.py
```diff
@@ -15,36 +15,51 @@
...
-    log.debug('Running command "%s"', " ".join(cmd))
+    log.debug(f"Running command [timeout={timeout}]: " + " ".join(cmd))
 
     # run using Popen
     proc = subprocess.Popen(
         cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs
     )
-    stdout_val, stderr_val = proc.communicate()
-    retval = proc.returncode
+    if timeout is None:
```
ljgray, line 52: I don't think we actually have to check for `timeout` being `None` ourselves, the default value for `timeout` in `.communicate` is `None` so we can just pass that along and the behaviour won't change. (i.e., just keep the `try: except:` block)

tests/conftest.py
```diff
@@ -53,6 +59,58 @@ def set_config(request):
...
 
+@pytest.fixture
+def mock_run_command(request):
+    """Mock alpenhorn.util.run_command to _not_ run a command.
+
+    The value returned by run_command() can be set by the test via the
+    run_command_result mark.
+
+    This fixture yields a function which returns a dictionary containing
+    the agruments passed to run_command.
```
ljgray, line 70:
```suggestion
    the arguments passed to run_command.
```

ljgray (changes requested):

alpenhorn/util.py
ketiltrout: Oooo, nice.

ketiltrout:
> > There are two constants defined in this file, `PULL_TIMEOUT_BASE` and `PULL_BYTES_PER_SECOND` which determine the length of time before the transfer is killed. I've indicated that these could be put in the alpenhorn config, but I'm not sure it's worth it. I'm looking for opinions.
> 
> Personally I lean towards having them in config, because if we ever end up wanting to change them for some reason it's more of a pain to hunt down where they're defined. Also, depending on how the deployment is done, a change would just have to involve redeploying the config (assuming it's an ansible play)

Done

ljgray (approved):

PR #153, opened 2023-03-02
Rewrite 10/14: DefaultIO file pulls

### LINKED ISSUE #48: Copy request fails if `node_from` has a NULL `address`
Line 304 in [`update.py`](https://github.com/radiocosmology/alpenhorn/blob/master/alpenhorn/update.py#L304) is filtering out my list of archive file copy requests to nothing:

```py
    requests = requests.join(st.StorageNode).where(st.StorageNode.address != 'HPSS')
```

It's probably because the request's `node_from.address` is NULL. When I remounted the node with an explicit `--address` argument, the copy went ahead fine.

ketiltrout:
This PR completes the re-write of `update.py` to transition to the new I/O framework by re-implementing support for pull requests.

## The UpDownLock
This is a fairly simple thread synchronization primitive that I'm a little surprised it doesn't already exist.

It's used to prevent race conditions during the modifications of a node directory tree by only allowing either creation or deletion to happen at any given time.  It's needed to prevent the following (starting from a node with only one file on it, `/node/acq/file1`:
```
    worker1 (pulling)                       worker2 (deleting)
    -----------------------                 -----------------------
    wants to create:                        wants to delete:
       /node/acq/file2                        /node/acq/file1
    -----------------------                 -----------------------
1.  verifies that /node/acq exists          deletes /node/acq/file1

2.  skips directory creation                deletes /node/acq (empty dir)

3.  tries to create /node/acq/file2                  ...

4.  catches on fire because /node/acq                ...
      is now missing after verifying
      it existed
```

Any number of threads may hold the lock in a particular state ("lock up" or "lock down") at any given time, but switching from one state to the other forces threads to wait for the lock to become unlocked first.

## Group I/O
Finally, the reason for having an I/O class for StorageGroups:  pull requests, because they target a group are now handled by the group I/O class. The job of the group I/O class is to:
* figure out if the file already exists in this group (`DefaultGroupIO.exists`) and, if it does, cancel the request
* otherwise, figure out which node in the group (if any) will receive the pull request (`DefaultGroupIO.pull`)

If the group selects a node for the request, the rest of the pull is then handed off to the node's I/O layer.

## Node I/O
The guts of the pull request haven't changed and live in the `pull_async`, which is just a cleaned-up version of what used to be in `update_node_requests`.

One wrinkle:
* because all the pull requests now (potentially) happen simultaneously, there needs to be a mechanism for avoiding pulling too much data to the node at once.  This is done via the `reserve_bytes` mechanism in DefaultIO which synchronously (i.e. in the main thread) tries to "reserve" a certain amount of the free space on the node before creating a pull task.  If the reservation fails, the request is skipped (node too full), until the next time through the loop when there might be space.
* There's a "reservation factor" which I've set to 2 (though the correct factor might be something else), meaning twice as much space is reserved than is needed.  This guarantees we'll never overfill a node as a result of actual-apparent size differences, but does mean a node will never get to 100% full.  Not sure if there's a better way to do this.

## Remote I/O
This PR also introduces a somewhat-awkward "Remote I/O" module which is used by alpenhorn to learn things about a non-local node (Technically, it's any node that's the source-side of a pull request, which may in fact be local).  The reason for this is knowledge about a remote node needs to be laundered through the I/O layer because the I/O class and I/O config may affect the information.

There's only two bits of information that alpenhorn needs from a remote node:
* what is the path to the remote file (`file_addr` or `file_path`) that can be given to bbcp/rsync
* is the remote file ready to be pulled (`pull_ready`)

This information could definitely just be part of the Node I/O class, but making it a separate object prevents trying to perform I/O operations on the source node of a pull request.

## Changes to pulls
I've made a few, hopefully minor, changes to the details of how a pull request happens:
* hard links are now never made between archive nodes and non-archive nodes.  I think this is a good idea for the purposes of data integrity of the archived data.  It prevents the accidental corruption of archived files which are accessed via a different node (where you might otherwise think corruption isn't a big deal because you can always just re-sync from the archive copy).
* I'm more explicit about what to do when there's a copy of a file on the pull destination (which may or may not be registered in the database).  Alpenhorn-1 would just fall over and complain when this happened.  Alpenhorn-2 before this rewrite was better about things: it just explicitly clobbered a destination file.  But now the response is more nuanced:
   * If there's an `ArchiveFileCopy` on the destination node with `has_file=='Y'`, then alpenhorn says "Job's a good'un" and cancels the request it was working on
   * If there's an `ArchiveFileCopy` with `has_file=='X'`, then the pull goes ahead and the corrupt file is overwritten.
   * If there's an `ArchiveFileCopy` with `has_file=='M'`, then alpenhorn figures the destination file is going to be checked for corruption soon, and just abandons the request without resolving it (reasoning that next time round, when we handle the request again, the check will have completed and changed `has_file` to "Y" or "X" as appropriate.)
   * Otherwise (`ArchiveFileCopy.has_file=='N'` or no copy record at all), then it looks for a file with the destination path.  If it finds one, it creates/updates an `ArchiveFileCopy` for the existing file and sets `has_file='M'` so that the unexpected file on the destination gets checked.  The request is in this case abandonned without resolution and, like the other `has_file=='M'` case above, it will come back to finish the request after the check is completed.
   * Ultimately, if none of the above happens, then the pull is clear to proceed.

## Broken by this PR
This PR deletes all the special casing previously in `update.py` for transport nodes.  A replacement, in the form of a Transport Group I/O class, will be implemented to fix this in the next PR.

## Resolved by this PR
I think this closes #48 in that a NULL `address` is only going to be a problem in cases when a pull request actually needs it now.

### DIFF
--- a/alpenhorn/io/_default_asyncs.py
+++ b/alpenhorn/io/_default_asyncs.py
@@ -2,17 +2,165 @@
 from __future__ import annotations
 from typing import TYPE_CHECKING
 
+import time
 import errno
+import shutil
 import logging
+import pathlib
 from datetime import datetime
 
-from ..archive import ArchiveFileCopy
+from . import ioutil
+from ..archive import ArchiveFileCopy, ArchiveFileCopyRequest
+from ..update import RemoteNode
 
 if TYPE_CHECKING:
     from .base import BaseNodeIO
+    from .updownlock import UpDownLock
     from ..task import Task
 
 log = logging.getLogger(__name__)
+
+
+def pull_async(
+    task: Task, io: BaseNodeIO, tree_lock: UpDownLock, req: ArchiveFileCopyRequest
+) -> None:
+    """Fulfill `req` by pulling a file onto the local node.
+
+    Things to try:
+        - hard link (for nodes on the same filesystem)
+        - bbcp (if source is not on this host)
+        - rsync if all else fails
+
+    Parameters
+    ----------
+    task : Task
+        The task instance containing this async.
+    io : Node I/O instance
+        The I/O instance for the pull destination node.
+    tree_lock : UpDownLock
+        The directory tree modificiation lock.
+    req : ArchiveFileCopyRequest
+        The request we're fulfilling.
+    """
+
+    # Before we were queued, NodeIO reserved space for this file.
+    # Automatically release bytes on task completion
+    task.on_cleanup(io.release_bytes, args=(req.file.size_b,))
+
+    # We know dest is local, so if source is too, this is a local transfer
+    local = req.node_from.local
+
+    # The Remote Node
+    remote = RemoteNode(req.node_from)
+
+    # Source spec
+    if local:
+        from_path = remote.io.file_path(req.file)
+    else:
+        try:
+            from_path = remote.io.file_addr(req.file)
+        except ValueError:
+            log.warning(
+                f"Skipping request for {req.file.path} "
+                f"due to unconfigured route to host for node {req.node_from.name}."
+            )
+            return
+
+    to_file = pathlib.Path(io.node.root, req.file.path)
+    to_dir = to_file.parent
+
+    # Placeholder file
+    placeholder = pathlib.Path(to_dir, f".{to_file.name}.placeholder")
+
+    # Create directories.  This must be done while locking up the tree lock
+    with tree_lock.up:
+        if not to_dir.exists():
+            log.info(f'Creating directory "{to_dir}".')
+            to_dir.mkdir(parents=True)
+
+        # If the file doesn't exist, create a placeholder so we can release
+        # the tree lock without having to wait for the transfer to complete
+        if not to_file.exists():
+            placeholder.touch(mode=0o600, exist_ok=True)
+
+    # Giddy up!
+    start_time = time.time()
+    log.info(f'Transferring file "{req.file.path}":')
+
+    # Attempt to transfer the file. Each of the methods below needs to return
+    # a dict with required key:
+    #  - ret : integer
+    #        return code (0 == success)
+    # optional keys:
+    #  - md5sum : string or True
+    #        If True, the sum is guaranteed to be right; otherwise, it's a
+    #        md5sum to check against the source.  Must be present if ret == 0
+    #  - stderr : string
+    #        if given, printed to the log when ret != 0
+    #  - check_src : bool
+    #        if given and False, the source file will _not_ be marked suspect
+    #        when ret != 0; otherwise, a failure results in a source check
+
+    # First we need to check if we are copying over the network
+    if not local:
+        if shutil.which("bbcp") is not None:
+            # First try bbcp which is a fast multistream transfer tool. bbcp can
+            # calculate the md5 hash as it goes, so we'll do that to save doing
+            # it at the end.
+            ioresult = ioutil.bbcp(from_path, to_dir, req.file.size_b)
+        elif shutil.which("rsync") is not None:
+            # Next try rsync over ssh.
+            ioresult = ioutil.rsync(from_path, to_dir, req.file.size_b, local)
+        else:
+            # We have no idea how to transfer the file...
+            log.error("No commands available to complete remote pull.")
+            ioresult = {"ret": -1, "check_src": False}
+
+    else:
+        # Okay, great we're just doing a local transfer.
+
+        # First try to just hard link the file. This will only work if we
+        # are on the same filesystem.  If it didn't work, ioresult will be None
+        #
+        # But don't do this if it creates a hardlink between an archive node and
+        # a non-archive node
+        if req.node_from.archive == io.node.archive:
+            ioresult = ioutil.hardlink(from_path, to_dir, req.file.name)
+        else:
+            ioresult = None
+
+        # If we couldn't just link the file, try copying it with rsync.
+        if ioresult is None:
+            if shutil.which("rsync") is not None:
+                ioresult = ioutil.rsync(from_path, to_dir, req.file.size_b, local)
+            else:
+                log.error("No commands available to complete local pull.")
+                ioresult = {"ret": -1, "check_src": False}
+
+    # Delete the placeholder, if we created it
+    placeholder.unlink(missing_ok=True)
+
+    if not ioutil.copy_request_done(
+        req,
+        io,
+        check_src=ioresult.get("check_src", True),
+        md5ok=ioresult.get("md5sum", None),
+        start_time=start_time,
+        stderr=ioresult.get("stderr", None),
+        success=(ioresult["ret"] == 0),
+    ):
+        # Remove file, on error
+        try:
+            to_file.unlink(missing_ok=True)
+        except OSError as e:
+            log.error(f"Error removing corrupt file {to_file}: {e}")
+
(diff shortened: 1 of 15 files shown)


alpenhorn/io/updownlock.py
```diff
@@ -0,0 +1,267 @@
...
+        # Otherwise, wait until unlocked
+        with self._is_unlocked:
+            if timeout < 0:
+                self._is_unlocked.wait()
+            else:
+                end_at = None
+                notified = False
+
+                # Wait until either we're notified or else we timeout
+                while not notified and end_at is None or time.monotonic() < end_at:
```
ketiltrout, line 100:
```suggestion
                while not notified and (end_at is None or time.monotonic() < end_at):
```
In Python, `and` binds more tightly than `or`.

alpenhorn/update.py
```diff
@@ -349,8 +358,12 @@ def update(self) -> None:
             self.update_delete()
 
             # Process any regular transfers requests onto this node
```
ketiltrout, line 360:
```suggestion
            # Prepare files for pulls out of this node
```

alpenhorn/io/_default_asyncs.py
```diff
@@ -2,19 +2,168 @@
...
+        check_src=ioresult.get("check_src", True),
+        md5ok=ioresult.get("md5sum", None),
+        start_time=start_time,
+        stderr=ioresult.get("stderr", None),
+        success=(ioresult["ret"] == 0),
+    ):
+        # Remove file, on error
+        try:
+            if to_file.exists():
+                to_file.unlink()
```
ketiltrout, line 155:
```suggestion
            to_file.unlink(missing_ok=True)
```
avoiding the (admittedly unlikely) race condition

alpenhorn/io/default.py
```diff
@@ -163,6 +218,109 @@ def md5(self, path: str | pathlib.Path, *segments) -> str:
...
+        ----------
+        req : ArchiveFileCopyRequest
+            the copy request to fulfill.  We are the destination node (i.e.
+            `req.group_to == self.node.group`).
+        """
+
+        if self.node.under_min:
+            log.info(
+                f"Skipping pull for StorageNode f{self.node.name}: hit minimum free space: "
+                f"({self.node.avail_gb:.2f} GiB < {self.node.min_avail_gb}:.2f GiB)"
```
ljgray, line 236:
```suggestion
                f"({self.node.avail_gb:.2f} GiB < {self.node.min_avail_gb:.2f} GiB)"
```

alpenhorn/io/default.py
```diff
@@ -163,6 +218,109 @@ def md5(self, path: str | pathlib.Path, *segments) -> str:
...
+            )
+            return
+
+        if self.node.check_over_max():
+            log.info(
+                f"Skipping pull for StorageNode f{self.node.name}: node full. "
+                f"({self.node.get_total_gb():.2f} GiB >= {self.node.max_total_gb}:.2f GiB)"
```
ljgray, line 243:
```suggestion
                f"({self.node.get_total_gb():.2f} GiB >= {self.node.max_total_gb:.2f} GiB)"
```

alpenhorn/io/default.py
```diff
@@ -163,6 +218,109 @@ def md5(self, path: str | pathlib.Path, *segments) -> str:
...
+            )
+            return
+
+        # Check that there is enough space available (and reserve what we need)
+        if not self.reserve_bytes(req.file.size_b):
```
ljgray, line 248: I might not be understanding correctly but aren't these space checks somewhat redundant? I would think that just checking `self.reserve_bytes` would be enough. At the very least, if `under_min` passes then `over_max` would also pass, so I'm not sure if both are needed. Or are the checks just done for proper logging of the issue?

ljgray (changes requested):

tests/io/test_defaultnode.py
```diff
@@ -46,3 +57,31 @@ def test_md5(unode, xfs):
...
+    # We start off with 10,000 bytes available
+    assert unode.io.reserve_bytes(4000, check_only=True) is True
+    assert unode.io.reserve_bytes(40000) is False
+
+    # Now reserve some bytes
+    assert unode.io.reserve_bytes(30000) is False
+
+    assert unode.io.reserve_bytes(2000) is True  # 4k bytes reserved
+    assert unode.io.reserve_bytes(2000) is True  # 8k bytes reserved
+    assert unode.io.reserve_bytes(2000) is False  # can't reserve 12k bytes
```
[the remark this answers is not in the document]
ketiltrout: I'll make the comments more explicit.

There's a factor of two when calling `reserve_bytes`:
https://github.com/radiocosmology/alpenhorn/blob/c4ead5d57b3ebcb7d0a5b32e03c1083b2aa4ee9d/alpenhorn/io/default.py#L264
which is present to avoid having to figure out how much space on disk a file of a given size is actually going to use up.  In this test, we start with 10,000 bytes free/unreserved.
* The first call reserves 4000 bytes (2 x 2000), leaving 6000 unreserved.
* The second call reserves another 4000 bytes (2 x 2000), leaving 2000 unreserved.
* The third call wants to reserve 4000 bytes again, but as there's only 2000 left unreserved at this point, it fails.

The factor of two is a _very_ conservative number, but I don't it matters too much: a factor of two results wastage of one file's worth of space on the device.

alpenhorn/io/default.py
ketiltrout: The tests are inter-related, but I don't think they're actually redundant.

The `over_max` and `under_min` tests are _database-only_ checks, which are not at all affected by the size of the file being transfered. They return the results of:
 * over max: `StorageNode.max_total_gb >= StorageNode.avail_gb`
 * under min: `StorageNode.min_avail_gb <= StorageNode.avail_gb`

The value of `avail_gb` _does_ ultimately get updated by checking the filesystem, but not during the check itself. 

The values of `max_total_gb` and `min_avail_gb` are set manually (when desired) to adjust how alpenhorn works:
* `max_total_gb` prevents a StorageNode from taking up more space than allotted to it.  We use it, for instance, on the `cedar_online` node to prevent the raw data from using all of our `/project` space.
* `min_avail_gb` prevents a StorageNode from using up all available space on a disk.  We could use it, for instance to ensure some minimum amount of space remains free on `/project`.

(On a filesystem which is used _only_ as a storage node, there _is_ a degeneracy here: setting `max_total_gb` to a value, _V_ is equivalent to setting `min_avail_gb` to: _(size-of-filesystem) - V_.)

Neither of these parameters need to be set (and, for most nodes they aren't), in which case these two checks automatically succeed.

Conversely, while `over_min` and `under_max` are database checks, the `reserve_bytes` check is an check against the filesystem (an I/O check).  It is meant to, regardless of database constraints we've put on a storage node, determine whether a given file is going to fit on the underlying filesystem.  (In cases where a single file is larger than the value of `min_avail_gb`, the `under_min` check can succeed while reservation can fail.)

The reserve bytes check is also necessary to prevent multiple pull jobs overfilling the filesystem by each job thinking there's enough space by not considering all the other pull jobs concurrently happening.

ljgray: That clears it up, thanks!

ljgray (approved):
Approved pending the remaining f-string fixes

PR #154, opened 2023-03-02
Rewrite 11/14: Transport Group I/O

ketiltrout:
This PR adds a new group I/O class, `TransportGroupIO` suitable for use with a transport group to fix the removal of transport group logic from `update.py` in the previous PR.

Features of a Transport StorageGroup:
* it may have any number of nodes.  All node must have `node.storage_type == 'T'`, but no restrictions are put on the I/O class of nodes.  Nodes in the group which have a different `storage_type` are ignored.
* all pulls to the StorageGroup must be local: non-local pull requests will be ignored
* when handling pull requests, transport nodes are prioritised by increasing free space: the group will attempt to pull a file to the fullest node that it thinks it will fit on.

The name of the group is no longer special (previously, the group had to be named "transport") and more than one independent transport group can be made now, though I don't know how useful that would be.

### DIFF
--- a/alpenhorn/io/transport.py
+++ b/alpenhorn/io/transport.py
@@ -1,0 +1,165 @@
+"""Transport Group I/O."""
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+import logging
+
+from .base import BaseGroupIO
+
+if TYPE_CHECKING:
+    import pathlib
+    from ..archive import ArchiveFileCopyRequest
+    from ..update import UpdateableNode
+
+log = logging.getLogger(__name__)
+
+
+class TransportGroupIO(BaseGroupIO):
+    """Transport Group I/O.
+
+    This implements (the formerly special-cased) transport disk logic.
+
+    A Transport StorageGroup is used to transfer data onto transiting
+    storage.
+
+    Features of a Transport StorageGroup:
+        - it may have any number of nodes.  All node must have
+            node.db.storage_type == 'T', but no restrictions are put on the
+            io_class of nodes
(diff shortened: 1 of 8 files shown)


alpenhorn/io/transport.py
```diff
@@ -0,0 +1,158 @@
...
+            the path to search for; relative to node root
+
+        Returns
+        -------
+        exists : UpdateableNode or None
+            If `path` is found on a node in the group, this is the node.
+            If `path` was not found, this is None.
+        """
+        for node in self._nodes:
+            if node.io.exists(path):
```
ljgray, line 87: Since the path is relative, couldn't this potentially be unstable if the same path exists on two nodes for whatever reason?

ljgray (changes requested):

alpenhorn/io/transport.py
ketiltrout: Yes, it will be unstable in that case, but my feeling is it shouldn't matter.

For an "ideal" transport group, it shouldn't be possible that there is more than one copy (because the file will be pulled at most once into the group), but in practice I don't think we should assume that happens.

The way I've designed it, if there are multiple copies, alpenhorn at least shouldn't make things worse by adding yet another copy.  It's going to pick one of those copies (via this method) to be the official one, and ignore the others.

It's also not going to automatically help you clean up such a mess by, say, deleting the extra copies.  Perhaps we could add something to the client to do that.

I'll add a note to the docstring indicating what's going to happen in this case.

ljgray (approved):

PR #156, opened 2023-03-04
Rewrite 13/14: Lustre I/O

ketiltrout:
This PR add the Lustre I/O classes.

## LFS wrapper
I've made a wrapper around calls to the `lfs(1)` binary in `alpenhorn.io.lfs.py`.  It is not a complete wrapper for that command and only able to run these commands:
* "lfs quota" to retrieve the group quota for a path.  The user group used must be specified in the I/O config.  Can also handle the quota max override needed for cedar's `/nearline`.
* "lfs hsm_state" to retrieve the HSM state for a file.  The state itself is represented with the `HSMState` enum and is one of:
  * `MISSING`: file is not present on disk or tape
  * `UNARCHIVED`: file exists on disk, but not on tape
  * `RESTORED`: file exists on both disk and tape
  * `RELEASED`: file exists on tape only
* "lfs hsm_restore" changes state `RELEASED -> RESTORED`
* "lfs hsm_release" changes state `RESTORED -> RELEASED`

## LustreQuota I/O

This is a very simple stepping-stone I/O module to Lustre HSM I/O.  It's exactly the same as DefaultIO, except is uses "lfs quota" to determine free space.  This behaviour re-creates the special casing we have in place in alpenhorn-1 for the "cedar_online" node.

## Lustre HSM I/O
The LustreHSM I/O is where all the HSM management occurs.  In general this is via updated versions of I/O methods from Default I/O. It also will automatically release files from disk (see `release_files`) as needed to keep enough headroom available on the HSM disk so that I/O can continue to happen.

The LustreHSM Group needs two nodes in it.  One of them must be a "LustreHSM" node, used as the "primary node", and another node, of any non-LustreHSM type, used as a "secondary node" for small files.  The group I/O config may specify the threshold for small files.

### DIFF
--- a/alpenhorn/io/lfs.py
+++ b/alpenhorn/io/lfs.py
@@ -1,0 +1,356 @@
+"""lfs(1) wrapper.
+
+This module provides the `LFS` class which wraps invocations of
+`lfs(1)` to interact with a Lustre filesystem.
+
+It is not a complete wrapper for that command and is only able to
+run these commands:
+
+* `lfs quota`
+    to retrieve the group quota for a path.  The user group used
+    must be provided to the constructor.  Can also handle a
+    fixed max quota, for cases where "lfs quota" doesn't correctly
+    report the max.
+* `lfs hsm_state`
+    to retrieve the HSM state for a file.  The state itself is
+    represented with the `HSMState` enum and is one of:
+        * `MISSING`:    file is not present on disk or external storage
+        * `UNARCHIVED`: file exists on disk, but not on external storage
+        * `RESTORED`:   file exists on both disk and external storage
+        * `RELEASED`:   file exists in external storage only
+* `lfs hsm_restore`
+    requests the state change `RELEASED -> RESTORED`
+* `lfs hsm_release`
+    requests the state change `RESTORED -> RELEASED`
+"""
+from __future__ import annotations
+from typing import TYPE_CHECKING
+
+import shutil
+import logging
+import pathlib
+from enum import Enum
+
+from alpenhorn import util
+
+if TYPE_CHECKING:
+    import os
+
+
+log = logging.getLogger(__name__)
+
+
+class HSMState(Enum):
+    """HSM States.
+
+    Indicates the state of a file in HSM (the external storage).
+
+    Four states are possible:
+    HSMState.MISSING:
+        The file is not on the system (disk or external storage) at all
+    HSMState.UNARCHIVED:
+        This file is on the Lustre disk but not in external storage.  This
+        is the state of newly created files until they are archived.
+    HSMState.RELEASED:
+        The file is in external storage but not on disk.
+    HSMState.RESTORED:
+        The file is both in external storage and on disk.
+
+    A new file starts off in state UNARCHIVED.  Initial archiving is beyond
+    our control, but once the file has been archived, it moves from state
+    UNARCHIVED to state RESTORED.
+
+    A `lfs.hsm_restore()` changes a file's state from RELEASED to RESTORED.
+    A `lfs.hsm_release()` changes a file's state from RESTORED to RELEASED.
+    """
+
+    MISSING = 0
+    UNARCHIVED = 1
+    RESTORED = 2
+    RELEASED = 3
+
+
+class LFS:
+    """A class that wraps invocations of the lfs(1) command for use on Lustre
+    filesystems.
+
+    If the lfs(1) command can't be found in the PATH, attempting to instantiate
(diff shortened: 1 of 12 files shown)


alpenhorn/io/nearline.py
```diff
@@ -0,0 +1,408 @@
+"""Cedar nearline I/O.
+
+These I/O classes are specific to the nearline tape archive at cedar.
```
ljgray, line 3: It's debatable, but I wonder if this should go in `alpenhorn-chime` instead?

alpenhorn/io/nearline.py
```diff
@@ -0,0 +1,408 @@
...
+            At least one node couldn't be identified, or more than
+            two nodes were provided.
+        """
+        if len(nodes) != 2:
+            raise ValueError(
+                f"need exactly two nodes in StorageGroup group {self.group.name} (have {len(nodes)})"
+            )
+            return False
+
+        if nodes[0].db.io_class == "Nearline":
```
ljgray, line 359: I'm assuming here that `smallfile` can be any io class?

ljgray (approved):

alpenhorn/io/nearline.py
ketiltrout: That's not a bad point...

alpenhorn/io/nearline.py
ketiltrout: Yeah, though probably should check that it's not also a Nearline node (since then there's no way to figure out which is which).

alpenhorn/io/nearline.py
ketiltrout: There's nothiing _explicity_ CHIME or cedar specific here; this should work on any HSM-backed lustre filesystem.  Maybe the thing to do is change this description.

alpenhorn/io/lfs.py
```diff
@@ -0,0 +1,286 @@
...
+        Returns
+        -------
+        restored : bool
+            True if a successful restore request was made, or if
+            `path` was already restored.  False otherwise.
+        """
+        state = self.hsm_state(path)
+
+        # If the file doesn't exist, fail
+        if state == HSMState.MISSING:
```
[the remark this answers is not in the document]
ketiltrout: Sure, I could add some debugging-level statements.

ketiltrout:
I've rewritten some docstrings.

It's good to have something that can test the file-latent parts of the I/O layer (i.e. ready/not ready).  There's nothing actually Cedar or CHIME specific here anymore except maybe the name.  Would it be better to rename this from `Nearline` to `LustreHSM`?

ketiltrout:
I've renamed:
* `Nearline` -> `LustreHSM`
* `LFSQuota` -> `LustreQuota` (just for consistency).

ljgray:
Changes look good. I noticed one small typo, line 163 in `auto_import`, in ddb9e40 to 3bfbecb which I must have missed when reviewing PR 12/14. It just says `copy.last_upate` instead of `copy.last_update`.

ketiltrout:
I think I added that after you reviewed it.  More importantly, though, is why the test suite didn't catch it.

ketiltrout:
Fixed and test-suite updated to catch such a thing

ljgray:
Looks great!

alpenhorn/io/lustrehsm.py
```diff
@@ -0,0 +1,426 @@
...
+        ------
+        ValueError
+            At least one node couldn't be identified, or more than
+            two nodes were provided.
+        """
+        if len(nodes) != 2:
+            raise ValueError(
+                f"need exactly two nodes in StorageGroup group {self.group.name} (have {len(nodes)})"
+            )
+            return False
```
ketiltrout, line 369: ?

ketiltrout:
I've obsessed a tiny bit over the output from `lfs quota` and ended up both fixing the parsing of the output from that command, and also making the following changes:
* `lfs.quota_remaining` now returns `None` if the group is using the default block quota setting but no `fixed_quota` has been provided.  Previously, in this case, `(0 - quota_used`) would be returned.
* in `LustreQuota`, when `fixed_quota` is needed to specify the default quota is specified
* in `LustreHSM` `fixed_quota` is no longer listed as required.
* in `LustreHSM`, `headroom` is no longer determined from `fixed_quota` if not specified, but is now listed as required.

Diff: https://github.com/radiocosmology/alpenhorn/compare/5bfa602..26696a292c3de6d9516d03cc03932ca54d79a467

PR #157, opened 2023-03-04
Rewrite 14/14: Idle updates

### LINKED ISSUE #75: Parallelize alpenhornd server
Currently, alpenhornd is single-threaded, despite the potential for long running operations fingerprinting and copying files. It seems this is trivially parallelizable at the file level. The rearchitecting needn't be large: the main loop could dispatch file operations to a worker thread pool that would run `auto_import._import_file()` and similar per-file functions.

(@jrs65 suggests per-storage node worker pool.)

### LINKED ISSUE #93: Port "transactions for upserting ArchiveFileCopy" from v1
PR 6 / commit 62a9d47

### LINKED ISSUE #136: Alpernhorn improvements for nearline intergration
@james-s-willis @ketiltrout here is a bit of a brain dump of things I think we want to do for this nearline work, which mostly summarises recent discussions we've had.

- Figure out a generalisation of transfers such that the source and destination can both participate. At the moment all transfers are pulls and so only the destination does anything, but for nearline transfers the alpenhorn at the source will need to ensure that the data is actually on spinning disk first. Questions: how to track this in the database (maybe add a column for `ready_source` that the source sets to true when it's done, and then the destination sets `completed`)? how to deal with transfers when a node doesn't have an alpenhornd managing it (which will currently work, but maybe shouldn't)? is there a way of doing this such that it is compatible with alpenhorn 1?
- Port alpenhorn 1 to use the new table schema, and make the changes to the database. Changes in the schema so far are:
  - `ArchiveAcq`: we removed the `inst` column in v2
  - `ArchiveFile`: we added a `registered` timestamp in v2
  - `StorageNode`: we renamed the `mounted` column to `active`
  - `ArchiveFileCopy`: we added a `size_b` column to track the actual on disk size
  - `ArchiveFileCopyRequest`: we changed it from one row for all equivalent transfers (with an `n_requests` column in v1) to a single row for each transfer in v2. We also added `transfer_started` and `transfer_completed` timestamps in v2
- Get alpenhorn 2 running against the current database. This may require getting the implementations of the CHIME specific tables finished that @jrs65 is working on
- Change the internals of alpenhorn 2 to have a async work queue for long running operations (source/destination end of transfers, md5sum's etc). The various `update_` methods should put items into this list rather than run them synchronously. Each item needs to encapsulate the operation it wants to do, perform any database changes, and any cleanup. This should probably be done with Python's async framework, although could probably be done with a queue and a threadpool. `asyncio.subprocess` will probably be very useful here.
- Add a way of having alpenhorn use different transport methods depending on the route. This might well take the form of a table of `node_from, node_to, source_op, dest_op, source_params, dest_params` with `source_op` and `dest_op` specifying operations being done at either end of the transfer, the params columns being maybe `JSONDict` types giving parameters that should be used.
- Alternatively this info above could be attached to nodes rather than pairs of nodes. That's less powerful, but certainly makes things like requiring a source op be done for every transfer from a node much easier. However, it makes it hard to describe things like that transfers cedar -> scinet must be pulls, but scinet -> cedar must be pushes due to the firewall rules.
- Implement a `source_op` that can perform a lustre hsm recall.

### LINKED ISSUE #140: TODO for running on cedar for CHIME
This is a tracking issue / to-do list for things that need doing before we can deploy this to cedar for CHIME.  There is no particular order for this list:

- [x] fix concurrency issues with #137
- [x] allow sub-node parallelisation
- [x] io-classing for IO abstraction
- [x] config blob for storage nodes
- [ ] ~node-group edge table~ (Probably not needed to get things running on cedar.)

See also: #136

ketiltrout:
This is the last PR of the rewrite.  It adds two task which are performed only when a node is idle (has no I/O tasks in the queue) during an update loop:
* On nearline nodes, check the HSM state of files and update the database.  Because the HSM state can be changed outside of alpenhorn, this is necessary to keep the DB in line with what's actually going on.
* On any node that enables it, automatically re-verify files to ensure they aren't corrupt.

## The QueryWalker

This PR introduces a `QueryWalker` in `querywalker.py` which facilitates both these tasks: it takes a table name (`model`) and (optionally) a bunch of condition expressions and walks through the matched rows every time `get()` is called.  The query starts at a random place in the results.  This is done that so short runs of alpenhorn won't always end up running over the same set of records.

## Config
For the HSM state check, the number of items to do at a time is given in the nearline I/O config.

For the re-verify check, this PR add a new field to the `StorageNode` model: `StorageNode.auto_verify` which is the number of files to auto-verify per idle update loop.  If this field is zero (the default), then auto-verification is not done.

## Note
Originally, this was just the HSM update thing, but then I realised the query-walker could also be used to implement this auto-re-verification, which I've been thinking about doing for a while, even though it's not really relevant to the rest of this rewrite

## Issues closed by the rewrite as a whole
Closes #75
Closes #93
Closes #136
Closes #140


alpenhorn/update.py
```diff
@@ -279,6 +338,10 @@ def update_idle(self) -> None:
             # Do any I/O class idle updates
             self.io.idle_update()
 
+            # Run auto-verfiy, if requested
```
ljgray, line 341:
```suggestion
            # Run auto-verify, if requested
```

alpenhorn/querywalker.py
```diff
@@ -0,0 +1,110 @@
+"""The QueryWalker class"""
+
+import peewee as pw
+from peewee import fn
+
+
+class QueryWalker:
+    """QueryWarlker class
```
ljgray, line 8:
```suggestion
    """QueryWalker class
```

alpenhorn/io/nearline.py
```diff
@@ -185,6 +266,92 @@ def check(self, copy: ArchiveFileCopy) -> None:
...
+
+        # Are we restored?  Note: we're interested in the actual
+        # state of the file, not the state as recorded in the database.
+        #
+        # If we are, we can just use the DefaultIO's check() method
+        # and not worry about any of the stuff going on below.
+        if not self._lfs.hsm_released(copy.path):
+            return super().check(copy)
+
+        def _async(
```
ljgray, line 311: What would happen if this is called on the same file multiple times? Since `QueryWalker` can return duplicates and this is asynchronous, it's entirely possible that we could call `auto_verify` on the same file twice

ljgray (dismissed):

alpenhorn/io/nearline.py
ketiltrout: Good point.

I think verification of the same file twice woudl work out fine.  It'd be a waste of effory (MD5-ing twice) but the last task running would end up winning.  Even if they overwrite each other, they should come to the same conclusion, so I think nothing too bad will happen.

Still, might be worth discarding duplicates from the list returned by the query walker.

However, that does make me think: what about the case when you turn on auto-verify on.a node with only one file?  Even if we do discard duplicates in a single update loop, alpenhorn's still just going to grind md5-ing that file over and over again for lack of anything else to do.  It might be worth adding a minimum wait time before re-verifying (i.e. skip everything with last_update less than a day (or week or something) old.

ljgray: Agreed, something like a day or two would probably be fine

ketiltrout: Both done:
* duplicates are skipped
* "new" files are not rechecked, where newness can be specified by the "auto_verify_min_days" config option and defaults to 7 days.  Because verification updates last_update, this will prevent re-verification from happening too often.

ketiltrout:
Updated to prevent over-verification

ljgray (approved):

PR #162, opened 2023-07-22
feat(logger): configure syslog and file logging

### LINKED ISSUE #161: Port log file handling code from alpenhorn1
Alpenhorn-1 uses https://pypi.org/project/concurrent-log-handler/ to handle log files, which is very helpful for us on the clusters where we can't use logrotate to handle log files.

We should re-implement this as an optional feature, configured via the config file (instead of the `ALPENHORN_LOG_FILE` environmental variable).  e.g.:
```
logging:
    level: DEBUG
    file:
        name: ~/alpenspool/alpenhorn.log
        rotate: yes
        max-size: 4G
        file-count: 100
```

ketiltrout:
Allows optional configuration to start up file or syslog logging.  File logging may include file rotation and/or watching.

Rotating file logging is necessary for running on cedar.  Log messages emitted before enabling the log file are buffered and flushed to the log file once it is opened, meaning the config-loading code can still use the log even before a log-file is opened.

Moves the initial logging setup from import time to an explicit call to `init_logging`.

Closes #161

### DIFF
--- a/alpenhorn/config.py
+++ b/alpenhorn/config.py
@@ -31,15 +31,77 @@ Example config:
         - alpenhorn_chime
         - chimedb.core.alpenhorn
 
-    # Logging configuration
+    # Logging configuration.  By default, alpenhorn sends all log message to
+    # standard error.
     logging:
-
         # Set the overall logging level
         level: debug
 
         # Allow overriding the level on a module by module basis
         module_levels:
             alpenhorn.db: info
+
+        # Alpenhorn can be configured to send log output to syslog and/or
+        # a file.  This is _in addition_ to the log sent to standard error,
+        # which is always enabled.
+
+        # Syslog logging.
+        syslog:
+            # If true, enable logging to syslog.  Note: if the "syslog" section
+            # is present in the logging config, then the default value
+            # of this key is true.  As a result, this key need only be specified if
+            # none of the other syslog configuration parameters are provided
+            # in the config.
+            enable: true
+
+            # The network address to send syslog message to.
+            #
+            # May also be a Unix domain socket (like "/dev/log").  In that case,
+            # set `port`, below, to zero to indicate this.
+            #
+            # Defaults to "localhost".
+            address: localhost
+
+            # The network port to send syslog messages to.  If this is zero,
+            # then the port is ignored and `address` is taken to be a Unix domain
+            # socket.  Defaults to 514, the standard syslog port.
+            port: 514
+
+            # The syslog facility to use.  If given, should be the name of one of
+            # the syslog facilities, disregarding case ("user", "local0", ...)
+            # Defaults to "user".
+            facility: user
+
+            # Set to True to use TCP, instead of UDP, to send messages to the syslog server.
+            # Ignored if using a Unix domain socket (i.e. `port` is zero).  Default is False.
+            use_tcp: false
+
+        # File logging.
+        file:
+            name: /path/to/file.log
+
+            # If a third-party (like logrotate) is rotating the alpenhorn log
+            # file, set this to "true" to tell alpenhorn to watch for log-file
+            # rotation.
+            watch: false
+
+            # Alternately, alpenhorn can manage log file rotation itself.  Set
+            # "rotate" to true to enable.  At most one of "watch" and "rotate"
+            # may be true.
+            rotate: true
+
+            # The following two settings affect alpenhorn-managed log rotation
+            # and are ignored if "rotate" is not true.
+
+            # Maximum number of rotated files to keep.  Must be at least one
+            # Rotated files append an integer to the name specified (e.g.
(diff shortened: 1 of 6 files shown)


alpenhorn/logger.py
```diff
@@ -1,51 +1,375 @@
...
 
-root_logger = logging.getLogger()
-
-log_stream = logging.StreamHandler()
+try:
+    from concurrent_log_handler import (
+        ConcurrentRotatingFileHandler as RotatingFileHandler,
+    )
+except ImportError:
+    RotatingFileHandler = logging.handlers.logging.handlers
```
ketiltrout, line 35:
```suggestion
    RotatingFileHandler = logging.handlers.RotatingFileHandler
```

alpenhorn/logger.py
```diff
@@ -1,51 +1,375 @@
...
+    # Format handler
+    global log_fmt
+    handler.setFormatter(log_fmt)
+
+    # Log the start of file logging to the alpenhorn logger.
+    # We do this _before_ adding the file handler to prevent
+    # duplicating this message (via both the file handler and
+    # the start-up handler).
+    alpen_logger = logging.getLogger("alpenhorn")
+    alpen_logger.info(f"Logging to{how} {name} == {handler}")
```
ketiltrout, line 308:
```suggestion
    alpen_logger.info(f"Logging to{how} {name}")
```

ketiltrout:
This PR introduces a dependency on `concurrent-log-handler`, which provides a version of the standard `RotatingFileHanlder` which allows multiple processes to write to the same log file.  This is done via a lock file.

This dependency was in alpenhorn1, which is why I've added it here, but I don't think it's necessary.  It's not clear to me why we would have multiple things writing to alpenhorn's log file at the same time.

Should I remove it?

The only thing I can think is that maybe we had it in alpenhorn1 due to the old HPSS callback script which used to be executed on another machine on scinet?  Even that is no longer done: the main alpenhorn process takes care of everything now for HPSS.

ljgray (approved):
I think this is fine. I don't really see a point in keeping `concurrent-log-handler` as a requirement if we don't need it, and since this is a full rewrite it's as good a time as any to remove it

alpenhorn/config.py
```diff
@@ -31,16 +31,77 @@
...
             alpenhorn.db: info
 
+        # Alpenhorn can be configured to send log output to syslog and/or
+        # a file.  This is _in addition_ to the log sent to standard error,
+        # which is always enabled.
+
+        # Syslog logging.
+        syslog:
+            # If true, enable logging to syslog.  Note: the default value
+            # of this key is true, and this key need only be specified if
```
ketiltrout, line 51:
```suggestion
            # If true, enable logging to syslog.  Note: if the "syslog" section
            # is present in the logging config, then the default value
            # of this key is true.  As a result, this key need only be specified if
```

alpenhorn/logger.py
```diff
@@ -1,51 +1,375 @@
...
+            super().close()
+        finally:
+            self.release()
+
+
+def init_logging() -> None:
+    """Initialise the logger.
+
+    This function is called before the config is read.  It sets up logging to
+    standard output and also starts a log buffer where messages accumulate
```
ketiltrout, line 109:
```suggestion
    standard error and also starts a log buffer where messages accumulate
```

alpenhorn/logger.py
```diff
@@ -1,51 +1,375 @@
...
+    # Log the start of syslogging to the alpenhorn logger.
+    # We do this _before_ adding the file handler to prevent
+    # duplicating this message (via both the syslog handler and
+    # the start-up handler).
+    alpen_logger = logging.getLogger("alpenhorn")
+    alpen_logger.info(
+        f"Logging to syslog at {address}:{port} via "
+        + ("TCP" if use_tcp else "UDP")
+        + " as facility {facname}"
+    )
```
ketiltrout, line 237:
```suggestion
    if port:
        alpen_logger.info(
            f"Logging to syslog at {address}:{port} via "
            + ("TCP" if use_tcp else "UDP")
            + f" as facility {facname}"
        )
    else:
        alpen_logger.info(
            f"Logging to syslog socket {address}"
            + f" as facility {facname}"
        )    
```

alpenhorn/logger.py
```diff
@@ -1,51 +1,375 @@
...
+    ---------
+    file_config
+        The contents of the `logging.file` section of the
+        alpenhorn config.
+
+    Returns
+    -------
+    file_handler
+        The configured file handler
+    """
```
ketiltrout, line 255:
```suggestion
    Raises
    ------
    KeyError
        no `logging.file.name` was specified in the config
    ValueError
        a bad value was encountered in the logging config
    """
```

PR #164, opened 2023-07-24
fix: Store/read (naive) UTC datetimes to/from DB.

### LINKED ISSUE #159: Database timestamps need to be UTC
Currently we store `datetime.datetime.now()` to the DB all over the place.  This gets confusing with servers spread across multiple timezones.

Dates stored to, and read from, the database should always be UTC times.

ketiltrout:
The `datetime`s are all naive because peewee doesn't work well with TZ-aware ones.

Closes #159

### DIFF
--- a/alpenhorn/acquisition.py
+++ b/alpenhorn/acquisition.py
@@ -40,7 +40,7 @@ class ArchiveFile(base_model):
     md5sum : string
         md5 checksum of file. Used for verifying integrity.
     registered : datetime
-        The time the file was registered in the database.
+        The UTC time when the file was registered in the database.
     """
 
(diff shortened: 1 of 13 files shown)


ljgray (approved):
I think it's a good idea. I guess that existing datasets won't chaneg to utc until they get updated for one reason or another, but if they're already in various timezones I don't see how that can make anything worse.

We would have to be careful making assumptions about whether or not a dataset time is utc (does that matter)?

ketiltrout:
This doesn't affect datasets or anything else in the database outside of the alpenhorn tables.

ljgray:
> This doesn't affect datasets or anything else in the database outside of the alpenhorn tables. Not even the times in the info tables.

Oh, duh

PR #168, opened 2023-08-18
Various small fixes found during the cedar shakedown

ketiltrout:
I don't think these are big enough to merit their own PRs:

# [fix(io): Add missing BaseGroupIO.set_storage()](https://github.com/radiocosmology/alpenhorn/commit/f7adfd0882186e9f440b090f1e488b24c04930c2)

Also:
* renamed BaseNodeIO.update() to BaserNodeIO.set_storage().
* enforce keyword arguments in UpdateableGroup.reinit().

The "set_storage" methods are called at the top of every update loop to replace the StorageNode/Group of the I/O classes. "update" was a very poor choice for their name because there are already too many things referred to as "update", and the name was always just going to lead to confusion.

(The erstwhile "update()" method was always called _before_ the "before_update()" method, which is a pretty clear indication there's something wrong with the name.)

# [Require peewee >= 3.16.3 to avoid deprecation warning](https://github.com/radiocosmology/alpenhorn/commit/bcced15248e209f3a092370f339e0cb5328e5a52)

Ported from https://github.com/chime-experiment/chimedb/pull/40

# [ Don't let pathlib.Paths leak into run_command ](https://github.com/radiocosmology/alpenhorn/commit/074f36b674e4b03d7f4ef431b8559f45c4bdd301)

`run_command` only accepts strings.

# [ Make INFO the default log level everywhere](https://github.com/radiocosmology/alpenhorn/commit/5cacc4e9e845d4eb54ad6bfc43d004757c58c6ab)

I get the idea of using `{"level": "warning", "module_levels": {"alpenhorn": "info"}}` as the logging config, but I think it's too confusing from a user point of view.  By default, users are going to expect the top-level `logging.level` to set the logging level of alpenhorn itself.  (Whereas, with the original config, it was setting the logging level of everything _but_ alpenhorn).

# [ Fix f-strings ](https://github.com/radiocosmology/alpenhorn/commit/787596ae3b9123ec9ffeedfeb0ccee39e687409e)

... which were missing their leading-`f`.

# [ Fix calcuation of StorageNode.get_total_gb() ](https://github.com/radiocosmology/alpenhorn/commit/c24045938646d4562b7cb8d023296fdecc121648)

It was returning MiB (`2**20`), instead of GiB (`2**30`).

### DIFF
--- a/alpenhorn/config.py
+++ b/alpenhorn/config.py
@@ -145,7 +145,7 @@ log = logging.getLogger(__name__)
 config = None
 
 _default_config = {
-    "logging": {"level": "warning", "module_levels": {"alpenhorn": "info"}},
+    "logging": {"level": "info"},
     "service": {
         "auto_import_interval": 30,
         "auto_verify_min_days": 7,
--- a/alpenhorn/db.py
+++ b/alpenhorn/db.py
@@ -189,7 +189,7 @@ class RetryOperationalError:
     See: https://github.com/coleifer/peewee/issues/1472
     """
 
-    def execute_sql(self, sql, params=None, commit=pw.SENTINEL):
+    def execute_sql(self, sql, params=None, commit=None):
         """Extend default execute_sql to retry.
 
         Retries once on pw.OperationalError, but only if not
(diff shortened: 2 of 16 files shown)


ljgray (approved):
All look file

PR #169, opened 2023-09-16
fix: Update ArchiveFileCopy.last_update as needed

ketiltrout:
I think these were overlooked because of my initial idea of making `ArchiveFileCopy.last_update` an automatically-updating value, which didn't pan out.

### DIFF
--- a/alpenhorn/io/lustrehsm.py
+++ b/alpenhorn/io/lustrehsm.py
@@ -141,9 +141,9 @@ class LustreHSMNodeIO(LustreQuotaNodeIO):
                 )
                 lfs.hsm_release(copy.path)
                 # Update copy record immediately
-                ArchiveFileCopy.update(ready=False).where(
-                    ArchiveFileCopy.id == copy.id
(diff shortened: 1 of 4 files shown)


ljgray (approved):
Changes look good, but what was the reason for _not_ making it an auto-updating value?

ketiltrout:
My original plan was to use the `ON UPDATE CURRENT_TIMESTAMP` clause, which is great, but only works in MySQL, which was one problem.

More importantly, though, was that the `peewee` data model doesn't really support automatically updating  fields.  I tried for a while to write an extension to `peewee.Model` which would handle it, but it turned out to be unnecessarily complex for what I wanted to do.

I briefly mentioned removed my auto-update attempt in https://github.com/radiocosmology/alpenhorn/pull/152

I think the primary downside to not having it auto-update is other processes that modify `ArchiveFileCopy` won't necessarily update it, but it's mostly a field to record when alpenhorn last checked the record, so it's probably fine.

PR #170, opened 2023-09-20
feat: Edge table and transfer race condition fixes

### LINKED ISSUE #158: Reasons to have a CHIME edge table
Since edge tables keep coming up, here some data to store in an edge table that CHIME could benefit from:
* **auto-transfer** flag: when new files appear on source node (at the end of a copy-request or import) queue AFCR to transfer to destination group
* **auto-delete** flag: if files are available on both source node and destination group, delete filecopies from source node

Essentially this would take over from all the transfer management we do with cronjobs right now.  It would also significantly reduce transfer latency, though I'm not sure how big a benefit that would have.

ketiltrout:
This PR handles a few potential race conditions in the way CHIME moves data around its Storage graph.  There are two parts.

# Change to deletion

The easier part is a change to the way alpenhorn collects files for deletion.  It now skips deleting a file copy which is needed to fulfill a copy request.  It doesn't cancel the deletion, so it will consider deleting it again next time, but deletion won't actually occur until the blocking copy request is handled (completed or cancelled), avoiding a potential race condition in file management.

# Edge table

The bigger change here is the addition of a table providing metadata for the edges in the Storage directed graph (edges in a directed graph are usually called "arrows").   This is the table `StorageTransfer` defined in `storage.py`.

We've talked about edge tables for a long time, and the PR implements almost nothing of what we've talked about, though there is potential to add more stuff later.

A `StorageTransfer` edge is defined by a `node_from` (source) and a `group_to` (destination).  Any `ArchiveFileCopyRequest` with the same `node_from` and `group_to` is transferring data along that edge, and there's the _potential_ to use the `StorageTransfer` to provide finer-grained configuration for specific routes.  But, as I alluded to above, none of that is implemented here.

What _is_ implemented is two post-file-adding actions which I've previously mentioned in #158 :
* **Autosync:** when a file is added to a node, autosync tells alpenhorn to immediately queue a new `ArchiveFileCopyRequest` to transfer it from this node to a destination group.  This allows us to define automatic transfer routes.  For example, we can turn on autosync on the `cedar_staging->scinet_staging` to automatically copy files to niagara when they arrive on cedar.  This feature is needed to avoid a race condition distributing files to `cedar_staging` and `cedar_offload` after transfer from gong.
* **Autoclean:** when a file is added to a node, autoclean tells alpenhorn to immediately mark the file copy on an "upstream" node for deletion (by setting `wants_file` to "N").  This feature isn't really needed to fix any problems, but, along with autosync it allows us to completely specify in the alpenhorn database our routine data movement using the following edge definitions (obviously, some of these won't work until we deploy alpenhorn2 to the other hosts):

| node_from | group_to | auto-sync | auto-clean | explanation
|-------|-------|-------|-------|-------|
| gong | cedar_staging | Y | N | files appearing on gong are automatically transferred to cedar_staging
| cedar_staging | cedar_offload | Y | N | files arriving on cedar_staging are automatically transferred to cedar_offload
| cedar_staging | scinet_staging | Y | N | files arriving on cedar_staging are automatically transferred to scinet_staging
| cedar_staging | scinet_hpss | N | Y | files are deleted from cedar_staging after being archived in HPSS on scinet
| cedar_offload | cedar_nearline | Y | Y | files arriving on cedar_offload are automatically transferred to nearline, and then deleted once they're in nearline
| scinet_staging | scinet_hpss | Y | Y | files arriving on scinet_staging are automatically transferred to HPSS, and then deleted once they're in HPSS

The post-add actions (autosync and autoclean) are triggered whenever a file appears on a node (i.e. both via import and also pull requests).

_Subtlety!_ The route a file takes **_does not matter_** when performing autosync and autoclean.  So, e.g., in the example table above, autocleaning of  `cedar_staging` will happen after files appear on HPSS, even though the `cedar_staging->scinet_hpss` edge is not used to transfer files into HPSS.

alpenhorn ignores all `StorageTransfer` records where `node_from.group == group_to` (i.e. edges pointing back to their origin, what are known as "self-loops" or 1-cycles).

Also: autoclean and autosync aren't complete replacments for cron-based `alpenhorn clean/sync` invocations.  These actions only ever trigger once, when a file first appears on a node.  It can't do a full sync/clean of a node.

Closes #158

### DIFF
--- a/alpenhorn/io/ioutil.py
+++ b/alpenhorn/io/ioutil.py
@@ -10,7 +10,7 @@ from datetime import datetime
 from tempfile import TemporaryDirectory
 
 from ..archive import ArchiveFileCopy, ArchiveFileCopyRequest
-from ..storage import StorageNode, StorageTransfer
+from ..storage import StorageNode, StorageTransferAction
 from .. import config, db, util
 
 if TYPE_CHECKING:
@@ -329,13 +329,13 @@ def post_add(node: StorageNode, file_: ArchiveFile) -> None:
         The file added.
     """
 
-    # Autosync: find all StorageTransfers where we're the source node
-    for edge in StorageTransfer.select().where(
-        StorageTransfer.node_from == node,
-        StorageTransfer.group_to != node.group,
-        StorageTransfer.autosync == True,
+    # Autosync: find all StorageTransferActions where we're the source node
+    for edge in StorageTransferAction.select().where(
+        StorageTransferAction.node_from == node,
+        StorageTransferAction.group_to != node.group,
+        StorageTransferAction.autosync == True,
     ):
-        if edge.group_to.filecopy_state(file_) == "N":
+        if edge.group_to.filecopy_state(file_) != "Y":
             log.debug(
                 f"Autosyncing {file_.path} from node {node.name} to group {edge.group_to.name}"
             )
@@ -344,24 +344,26 @@ def post_add(node: StorageNode, file_: ArchiveFile) -> None:
                 node_from=node, group_to=edge.group_to, file=file_
             )
 
-    # Autoclean: find all the StorageTransfers where we're in the dest group
-    for edge in StorageTransfer.select().where(
-        StorageTransfer.group_to == node.group,
-        StorageTransfer.node_from != node,
-        StorageTransfer.autoclean == True,
+    # Autoclean: find all the StorageTransferActions where we're in the
+    # destination group
+    for edge in StorageTransferAction.select().where(
+        StorageTransferAction.group_to == node.group,
+        StorageTransferAction.node_from != node,
+        StorageTransferAction.autoclean == True,
     ):
-        try:
-            copy = ArchiveFileCopy.get(
-                file=file_, node=edge.node_from, has_file="Y", wants_file="Y"
+        count = (
+            ArchiveFileCopy.update(wants_file="N", last_update=datetime.utcnow())
+            .where(
+                ArchiveFileCopy.file == file_,
+                ArchiveFileCopy.node == edge.node_from,
+                ArchiveFileCopy.has_file == "Y",
+                ArchiveFileCopy.wants_file == "Y",
             )
+            .execute()
+        )
 
+        if count > 0:
             log.debug(f"Autocleaning {file_.path} from node {edge.node_from.name}")
-
-            copy.wants_file = "N"
-            copy.last_update = datetime.utcnow()
-            copy.save()
-        except pw.DoesNotExist:
-            pass
 
 
 def copy_request_done(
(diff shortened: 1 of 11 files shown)


ljgray (approved):
I think this looks fine, but I'm not really up to date on the previous discussion about edge tables so it might be worth waiting for Richard to chime in

alpenhorn/storage.py
```diff
@@ -374,3 +374,60 @@ def update_avail_gb(self, new_avail: int | None) -> None:
 
         if new_avail is None:
             log.info(f'Unable to determine available space for "{self.name}".')
+
+
+class StorageTransfer(base_model):
```
jrs65, line 379: As I think about it more and more I think the naming for this is awkward. To me it sounds like it should represent a single transfer, i.e. a file being moved etc. In reality it represents configuration for what actions should be triggered when a file is moved. Maybe `StorageTransferRule`? `StorageTransferAction`?

For the future, I think if we want configuration for the actual transfers along specific edges, it's probably clearest to put them in a separate table.

alpenhorn/io/ioutil.py
```diff
@@ -314,6 +316,54 @@ def hardlink(from_path, to_dir, filename) -> dict | None:
...
+            )
+
+    # Autoclean: find all the StorageTransfers where we're in the dest group
+    for edge in StorageTransfer.select().where(
+        StorageTransfer.group_to == node.group,
+        StorageTransfer.node_from != node,
+        StorageTransfer.autoclean == True,
+    ):
+        try:
+            copy = ArchiveFileCopy.get(
```
jrs65, line 354: Could you just update all relevant file copies in here to save a query? Rather than the .get and .save combo.

jrs65 (changes requested):
This looks really great to me, and I think it'll simplify a lot of what we do.

In many ways I think it would be nicer if we were able to request what we actually want to happen, and give hints about how it should be done, e.g. we request files to go gong -> cedar_nearline, but specify that the route is through cedar_staging and then cedar_offload. I think that makes the intent of what we want to happen clearer, rather than it being done by a series of triggered requests.

*However*, that seems like it would add a lot of complexity in the implementation, particularly if you want to do it optimally (e.g. recognise that cedar_staging is on the the path to both cedar_offload and scinet_staging), and so I think this is still a massive improvement and it exists now.

My only real issue is that I find the naming of `StorageTransfer` kind of confusing.

ketiltrout:
In addition to implementing Richard's suggestions, I've changed autosync to fire whenever the destination has has_file!='Y' (instead of just when has_file=='N'/is missing).  I think this makes more sense.

In the case of `has_file=='M'`, the autogenerated pull request will not be acted upon on until the existing destination copy is checked.  After checking, if the resultant copy is set to `has_file=='Y'` alpenhorn will cancel the unnecessary pull request.  On the other hand, if the file ends up being corrupt (`has_file=='X'`), then the autosync will happen to overwrite the destination file.

jrs65 (approved):

PR #178, opened 2024-05-15
logging tweaks

ketiltrout:
Some changes to the logger output that I made a long time ago on cedar and deployed there but apparently never pushed down into master.


ljgray (approved):
Seems fine

rikvl (approved):

Issue #175, opened 2023-10-26
Zombies created by workers / wedged worker processes


ketiltrout:
During some Lustre I/O issues on cedar, we ran into the following zombie process:
```
chimedat 37016  1.3  0.0 1094268 111908 pts/13 Ssl+ Oct23  38:44  \_ /home/chimedat/alpenvenv/bin/python /home/chimedat/alpenvenv/bin/alpenhornd
chimedat 37305  0.0  0.0 212200 12292 ?        Ssl  Oct23   0:02      \_ orted --hnp --set-sid --report-uri 8 --singleton-died-pipe 9 -mca state_novm_select 1 -mca ess hnp -mca pmix ^s1,s2,cray,isolated
chimedat 39198  0.0  0.0      0     0 pts/13   Z+   Oct24   0:00      \_ [bbcp] <defunct>
```
which was not being reaped by its worker, meaning the worker never finished it's task.  This, in turn, prohibited further transfers into `cedar_staging` because the wedged worker was keeping the node busy.

We probably should have a mechanism for the main process to detect workers that have become wedged and deal with them.

Note: it's not clear to me what state the worker was in and why it wasn't able to reap the child.  Outside this one stuck worder, the remainder of alpenhornd was working normally.

ketiltrout:
There may be two somewhat separate ideas here, viz.:
* Tasks getting stuck and leaving nodes/groups busy.
* Child processes not being reaped and leaving zombies in the process table.

The solution to these might be two separate things.

PR #179, opened 2024-05-15
fix(run_command): fix subprocess timeout recovery

ketiltrout:
Normally, when a transfer times out, it's because it's stuck deep down in I/O wait, meaning SIGKILL
is not going to be handled in a timely manner, and, so, trying to `.communicate()` with it again is just futile.

With the current code, after a timeout, the worker just gets stuck in the same wait it tried to get out of with the kill, as if there were no timeout given.

So, let's just send it a kill and then abandon it.

Related to #175 (abandonning subprocesses may lead to zombies), but this doesn't really address the items brought up in that issue, which are higher-level ideas.

I have deployed this to cedar.

### DIFF
--- a/alpenhorn/util.py
+++ b/alpenhorn/util.py
@@ -30,6 +30,7 @@ def run_command(
     -------
     retval : int or None
         Return code, or None if the process was killed after timing out.
+        Integer zero indicates success.
     stdout : string
         Value of stdout.
     stderr : string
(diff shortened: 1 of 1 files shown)


rikvl (approved):
Ok.

I think (the newest version of) `black` wants to add empty lines and such in a bunch of files

ketiltrout:
It was the semicolons!  (I think I've been spending too much time in perl recently).

ljgray (approved):

PR #194, opened 2024-08-07
feat(queue): Scheduling for exclusive tasks

ketiltrout:
I think the solution for the problem outlined in #193 is to only run the "tidy-up" task when no other I/O is happening on the node.  This task is unusual in that it has the potential to modify anything on the node.  (Typically, tasks only concern themselves with a single file that they can be fairly certain no other tasks are also working on).

So here, I add the ability to flag an item in the task scheduler (i.e. the `FairMultiFIFOQueue`) as "exclusive".

* Exclusive items aren't returned by get() unless there's nothing else in-progress from its FIFO.
* When an exclusive item is returned by get(), the FMFQ locks its FIFO, preventing further items from that FIFO from being returned. The FIFO remains locked until the next task_done() call for that FIFO.

Before implementing this, I considered changing the way group idleness is computed, because that could fix it, too (the root problem is that the group is submitting tasks for the node, which adds a wrinkle to the update loop logic). I think this is probably cleaner, though I could be convinced otherwise, if that seems like a better solution.

### DIFF
--- a/alpenhorn/queue.py
+++ b/alpenhorn/queue.py
@@ -15,8 +15,8 @@ The queue is unbounded.
 
 import heapq
 import threading
-from time import time
 from typing import Any
+from time import monotonic, sleep
 from collections import deque
 from collections.abc import Hashable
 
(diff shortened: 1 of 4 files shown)


ljgray (approved):
I'm happy with this option. I think that the alternative (changing idle logic) would end up being more confusing to interpret and debug

PR #195, opened 2024-08-07
fix(io/default): Make the tidy-up task exclusive

ketiltrout:
The last part of the fix for the problem outlined in #193 and #194: make the tidy-up task exclusive.

Requires #194


ljgray (approved):

ketiltrout:
FYI: I've taken the opportunity in this PR to also reduce the frequency at which the tidy-up task runs by a factor of four.

For nodes that were typically idle (like `cedar_offload`), it was running roughly once every four hours, which seems too often.

Maybe this should be a parameter (either a ioconfig param, or a config-file param), but not here...

PR #200, opened 2024-10-29
fix(client): Client init.

### LINKED ISSUE #20: Refactor duplicate config+db+extension code into a single function
We have the same code for to load the config, start the database, and register extensions duplicated in a couple of places. (Right now it's in `alpenhorn.client` in `_init_config_db` and `init`, as well as `alpenhorn.service` in `cli`.)  Can we refactor it into a function in `alpenhorn/__init__`?

Leaving out the comments, this is the same set of steps that always (should? must?) happen:
```python
    config.load_config()
    extensions.load_extensions()
    db.config_connect()
    extensions.register_type_extensions()
```

ketiltrout:
The client's still quite broken, but at least it can start-up now...

This moves the common alpenhorn start-up from `server.service` to `common.util.start_alpenhorn`  (closes #20).

Then deletes `alpenhorn.client.connect_db` which was trying to do the same thing that `common.util.start_alpenhorn` is doing now.

Adds a common option to both client and server (--conf / -c) to allow specifying an additional config file on the command line.

Client logging has been sorted out.  It won't ever log to a log file, but verbosity can be adjusted via command-line.  Specific differences to client/server logging have been implemented, meaning a few parameters need now to be passed to `init_logging`.

<s>Requires (and currently targets) #199</s>


alpenhorn/common/logger.py
```diff
@@ -103,32 +132,83 @@ def close(self) -> None:
...
+    verbosity_to_level = {
+        1: logging.ERROR,
+        2: logging.WARNING,
+        3: logging.WARNING,
+        4: logging.INFO,
+        5: logging.DEBUG,
+    }
+
+    if verbosity not in verbosity_to_level:
+        raise ValueError("Bad verbosity: {verbosity}")
```
ketiltrout, line 161:
```suggestion
        raise ValueError(f"Bad verbosity: {verbosity}")
```

alpenhorn/common/config.py
```diff
@@ -194,21 +209,15 @@ def load_config():
             config = merge_dict_tree(config, conf)
 
     if not any_exist:
-        raise RuntimeError("No configuration files available.")
-
-
-class ConfigClass(object):
```
ketiltrout, line 200: I've deleted this vestigial thing that seemingly was never properly implemented.  (The `config` module was the one module that escaped the Great Refactoring or this probably would have been deleted then.)

ljgray (approved):

PR #201, opened 2024-10-30
fix(pyproject): Find subpackages

ketiltrout:
There's a lot of subpackages now.  And there will be more coming with the client re-write.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,15 @@ alpenhorn = "alpenhorn.client:cli"
 alpenhornd = "alpenhorn.server:cli"
 
 [tool.setuptools]
-packages = ["alpenhorn"]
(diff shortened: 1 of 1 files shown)


ljgray (approved):
This is fine, but I think you could just do something like 
```toml
[tool.setuptools.packages]
find = {}
```

like we've done in other repos, which should have the same effect

ketiltrout:
Yeah, thinking about it more, I think that's going to be more maintainable.  Let me change it...

ketiltrout:
Package installation has been broken since at least #185, but I think it might have been broken all the way back in the re-write.  We apparently never directly install it...

ljgray (approved):
Yeah that's what we've done everywhere else and it seems to have worked

Issue #202, opened 2024-10-31
Client command list


ketiltrout:
A proposal for the client command list, to see what we need.  Per #122, we will stick to `<noun> <verb>`:
* [x] `db` - database maintenance #231 
  * [x] `db init` - create tables, if necessary
  * Additional possibilities:  See #232 
    * `db status`.  Status of what?
    * `db check`.  Check integrity (table interconnections) 
* [x] `group` - for StorageGroup: Implementation #204 
  * [x] `group autosync GROUP NODE` :  #228
  * [x] `group create` - create new group
  * [x] `group list` - list groups
  * [x] `group modify GROUP` - edit group metadata
  * [x] `group nodes GROUP` - list nodes in group
  * [x] `group rename GROUP` - rename group
  * [x] `group show GROUP` - show StorageGroup record
    * [x] `--actions` - to show auto-actions:  #228
    * [x] `--transfers` - to summarise pending transfers in #238 
  * [x] `group sync GROUP NODE`  #233
    * An alias for `node sync NODE GROUP`
  * Additional possibilities:
    * `group add-node`.  See #205 
* [x] `node` - for StorageNode: Partial implementation #221 
  * [x] `node activate NODE` - activate node
  * [x] `node autoclean NODE GROUP` :  #228
  * [x] `node clean` - clean node of file copies #223
    * [x] `--include-bad` - also clean suspect/corrupt files.  #226 
      * See #225 
    * See #65
    * See #89
    * See #92
  * [x] `node create` - create new node.
    * See #39
  * [x] `node deactivate NODE` - deactivate node
  * <s>`node files NODE...` list files on node</s> See `file status`
  * [x] `node init` #237
    * See #227 
  * [x] `node list` - list nodes (curently called "status")
    * See #121
  * [x] `node modify NODE` - edit node metadata
  * [x] `node rename NODE` - rename node
  * [x] `node scan` - scan the node for known files to add (in alpenhorn-1 this was called "import-files", but in alpenhorn-2 client currently called "scan") #236 
  * [x] `node show NODE` - show StorageNode record: #224
    * [x] `--actions` - to show auto-actions:  #228
    * [x] `--transfers` - to summarise pending transfers out #238 
  * [x] `node stats NODE` - show file counts/sizes on node
  * [x] `node sync NODE GROUP` #233 
    * See #106
    * See #172
  * [x] `node verify NODE PATH...` re-verify file(s) on node. #229 
    * See #133
    * See #141
[...]

Issue #165, opened 2023-07-26
New parameters needing client support


ketiltrout:
The following parameters were added as part of the server rewrite and need to have support added to the client:

## StorageGroup
Implementation: #204
* [x] `StorageGroup.io_class`: should check whether the io_class is supported and warn (but still allow)
* [x] `StorageGroup.io_config`: needs JSON support.  Would be good to have the client vet the JSON and tell the user if required keys are missing and/or unknown keys are present for the `io_class` given.
  * the last part would require a way for the client to interrogate the I/O modules for the config metadata.

## StorageNode
Implementation #221 
* [x] `StorageNode.io_class` See StorageGroup, above
* [x] `StorageNode.io_config`: See StorageGroup, above
* [x] `StorageNode.auto_verify`: a non-negative integer

## ArchiveFileCopy
Implementation #259 
* [ ] `ArchiveFileCopy.ready`: generally the server should be able to figure this out by itself, but I don't see a problem with having a way of manually manipulating this in the rare case it's needed

Also, now:

## StorageTransferAction
* [x] all of it :  #228

PR #204, opened 2024-11-01
fix(client): Fix "group" command group and client test-suite

ketiltrout:
Here's a start on getting the client working, which I've done to try to gauge how much work it will entail.  This PR updates and re-implements command in the `group` command group, which manipulates StorageGroups.  (I think this is the easiest command group to do.)

## Test suite
This PR also re-engineers how the client is run in the test-suite.  There's now a `client` fixture that runs the client via `click.testing.CliRunner` and performs rudimentary checks on the result.  This abstracts a lot of the routine rigamarole of running the CLI via pytest, meaning the tests themselves can be relatively simple.

This system uses the same kind of persistent in-memory Sqlite database that I used for the server end-to-end test to make it possible for the client process and the test suite itself to access the same ephemeral database.

## Client
Common command options (of which there are many) are being collected in `client.options` along with common code to process them.

### Commands
(per the list in #202):

* `create` lets you set I/O class and I/O config.  There is no vetting of the I/O config data based on what the I/O class needs (see discussion in #165).  I think it's best to punt that downfield and worry about it in the future.
* `list` is pretty simple.  Shows I/O class but not I/O config, because I don't think it fits very well in the tabular format, but I could be convinced otherwise.
* `modify` has the same comments about I/O config vetting as `create`.  It can also fix broken JSON in the database.
* `rename` trivial
* `show` definitely willing to take suggestions on formatting here, as well as what StorageNode columns should be presented with `--node-details`.  `--node-stats` is for showing counts/sizes of files (see the `node stats` command in #202) and will be implemented when `node stats` is implemented.

## Also
Fixed `common.logger.echo` which was broken from the start.


ketiltrout:
In case it helps, here's typical output from `group list`:
```
$ alpenhorn group list
Name    I/O Class    Notes
------  -----------  -------
Group1  Default      Note1
Group2  Class2       Note2
```
and `group show`:
```
$ alpenhorn group show SGroup --node-details
Storage Group: SGroup
        Notes: Comment
    I/O Class: IOClass

I/O Config:

  Param1: 1
  Param2: 2

Nodes:

Name    Host        Active    I/O Class
------  ----------  --------  -----------
Node1   over_here   Yes       Default
Node2   over_there  No        NodeClass
```
(These are composed from several tests in the test suite).

ljgray (approved):

PR #212, opened 2024-11-06
fix(db): Fix db.threadsafe

ketiltrout:
Something I found while testing #210:

If we want db._base.threadsafe to be imported into `__init__`, it can't be a scalar (because it's going to change after the import).

Instead, make it a function.

Bonus: no need to try to import `subprocess` every time a command is run.  Just import it once at module import.


ljgray (approved):

ketiltrout:
This is running on cedar already as part of the dirty clone I'm using to test #210 .  I'm going to leave it there as-is and rebase back to master once #210 is tested and debugged and merged.

PR #218, opened 2024-11-09
feat(server): More hedging against I/O lock-up

ketiltrout:
If nothing else, I've managed to find a lot of places where iowait can lock up alpenhorn, I suppose...

This latest round tires to take care of the latest problem where all the workers were stalled while trying to check suspect files on nearline.

So, here's some more I/O-wait mitigation down in the `check` internals:
* Added a timeout directly to `alpenhorn.common.util.md5sum_file`.

  Uses `asyncio` to return early with a timeout error if it takes more than 10 minutes to MD5sum any 32 MiB chunk of the file, which seems like a very conservative timeout, but I'm willing to listen to other options.

* Wrapped the size-finding `stat` in `io._default_asyncs.check_async` in an asyncio timeout block for the same reason.

* Finally, replaced another instance of existance-`stat`-ting in `LustreHSMNodeIO.check._async` with a `lfs hsm_stat` invoke via the `LustreHSMNodeIO.exists` method.

I've put 10-minute timeouts all over the place, because I can't come up with a better number.  None of these operations should take anywhere near this long, but, on the other hand, 10 minutes is a fairly short time period in alpenhorn-server terms.

We're running from this branch on cedar.


rikvl (approved):
I don't see any problems.

Since the 10 min occurs in three different places (though maybe for slightly different purposes), do you maybe want to put it in a variable / configuration file? Would `_default_config` in `common/config.py` be the place for that?

PR #219, opened 2024-11-10
feat(server): A few optimations for deep queues

ketiltrout:
A couple of changes to reduce the amount of churn when it's taking a long time between job submission and execution because of a lot of other stuff already in the queue:

* when looking for file to ready for a remote pull, don't (re-)submit a ready job for things that are already ready (which is harmless to do, but not in any way helpful

* at the start of a pull job, re-check whether the file copy already exists on the destination node.  Again: would be harmless not to check: a new copy of the same file would overwrite the old one, but there's no harm to checking.

  This takes care of the case when two pulls for the same file happen in series.  It does not solve the case of the two identical pulls occurring in parallel.

### DIFF
--- a/alpenhorn/io/_default_asyncs.py
+++ b/alpenhorn/io/_default_asyncs.py
@@ -45,6 +45,18 @@ def pull_async(
         The request we're fulfilling.
     """
 
+    # Recheck the database to see if a file hasn't shown
+    # up somehow
+    if io.node.filecopy_state(req.file) == "Y":
+        log.info(
+            f"Cancelling pull request for {req.file.path} "
+            f"to {io.node.name}: already present."
+        )
+        ArchiveFileCopyRequest.update(cancelled=1).where(
+            ArchiveFileCopyRequest.id == req.id
+        ).execute()
+        return
+
     # Before we were queued, NodeIO reserved space for this file.
     # Automatically release bytes on task completion
     task.on_cleanup(io.release_bytes, args=(req.file.size_b,))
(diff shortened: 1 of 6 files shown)


alpenhorn/io/_default_asyncs.py
```diff
@@ -45,6 +45,18 @@ def pull_async(
         The request we're fulfilling.
     """
 
+    # Recheck the database to see if a file hasn't shown
+    # up somehow
+    if io.node.filecopy_state(req.file) == "Y":
+        log.info(
+            f"Cancelling pull request for {req.file.path}: "
```
rikvl, line 52:
```suggestion
            f"Cancelling pull request for {req.file.path} "
```
Not sure if that colon was meant to be there

alpenhorn/io/ioutil.py
```diff
@@ -427,15 +427,16 @@ def copy_request_done(
             stderr = "Unspecified error."
         if check_src:
             # If the copy didn't work, then the remote file may be corrupted.
-            log.error(f"Copy failed: {stderr};  Marking source file suspect.")
+            log.error(f"Copy failed:.  Marking source file suspect.")
```
rikvl, line 430:
```suggestion
            log.error(f"Copy failed.  Marking source file suspect.")
```
Same here

tests/server/test_update_node.py
```diff
@@ -393,22 +393,40 @@ def test_update_node_run(
 
     # And something to ready for a pull
```
rikvl, line 394:
```suggestion
    # And something ready for a pull
```

rikvl (approved):
Only some punctuation comments.

PR #220, opened 2024-11-10
refactor: Use "daemon" and "cli" instead of "server/client"

ketiltrout:
Talking to the CGEM folks, I think our (my) use of the terms "server" to refer to the long-running process that manages alpenhorn nodes, and "client" to refer to the command-line interface to the data index are unnecessarily confusing.

I want to move to using "daemon" and "cli" for these pieces of alpenhorn instead.

The terms "server" and "client" imply that the two pieces of alpenhorn are a normal server-client system, which they are not.  To over-simplify things, normally a "server" is a long-running process that produces data for users.  The users use the "client" to connect to the server and retrieve the data.

The two parts of alpenhorn do neither of these things.  The daemon is not at all user-facing.  It doesn't "serve" anything that anyone could retrieve.  It's concerned with managing files on disk (or tape).

Neither is the alpenhorn CLI a "client" in the traditional sense.  It doesn't interact with the server directly at all, and there's no API or protocol defined for communication between the two parts.  The CLI is a tool for reading and writing the data index.  By design, the two pieces can only communicate indrectly through the database.

AFAICT, the alpenhorn daemon was originally called a "daemon" (that's why it's `alpenhornd`.  Later this migrated to "service" (the alpenhorn-1 daemon is implemented in the module `alpenhorn.service`) and then later still to "server".  A lot of the "server" nomenclature shows up during my multithreaded rewrite of the daemon.  So, using "daemon" is something of a return to the original name.

The CLI has been referred to as both the "CLI" and the "client" interchangably since we started.  (alpenhorn-1's CLI is implemented in the module `alpenhorn.client`).  Here I want to push "CLI" over "client" to avoid slipping back into "client-server" terms in the future.

Though, I am willing to listen to other options for the names of these two pieces.  At some point we need to write some documentation for this system, and it would be good to have some concrete terms to work with for that.


rikvl (changes requested):
Nice! Thanks for the detailed explanation / motivation.

A search for `client` still finds a few instances of "client" or "client interface". Some of these could perhaps be changed to "command-line interface" or "CLI":
* https://github.com/radiocosmology/alpenhorn/blob/a0a12a561efe80ddc8b068555a656de7fb0f3d28/alpenhorn/cli/cli.py#L83
* https://github.com/radiocosmology/alpenhorn/blob/a0a12a561efe80ddc8b068555a656de7fb0f3d28/alpenhorn/cli/transport.py#L1
* https://github.com/radiocosmology/alpenhorn/blob/a0a12a561efe80ddc8b068555a656de7fb0f3d28/doc/index.rst?plain=1#L14
* https://github.com/radiocosmology/alpenhorn/blob/a0a12a561efe80ddc8b068555a656de7fb0f3d28/tests/cli/group/test_modify.py#L113
* Three in [`demo/demo-script.md`](https://github.com/radiocosmology/alpenhorn/blob/a0a12a561efe80ddc8b068555a656de7fb0f3d28/demo/demo-script.md)
* I presume the files `tests/cli/test_client*.py` don't matter since they seem be skipped.

There are also mentions of `service`. Not sure if you had in mind for those to stay or not. I think it would be less confusing if these would also change to "daemon", but perhaps I'm not following some subtleties here. For instance:
    * The filename of [`alpenhorn/daemon/service.py`](https://github.com/radiocosmology/alpenhorn/blob/a0a12a561efe80ddc8b068555a656de7fb0f3d28/alpenhorn/daemon/service.py)
    * The `service` section of the config and all references to it: https://github.com/radiocosmology/alpenhorn/blob/a0a12a561efe80ddc8b068555a656de7fb0f3d28/alpenhorn/common/config.py#L153

Finally, I found the function `deamon.service.cli` with the docstring `Alpenhorn data management service.` This got me thoroughly confused :) Is this a separate CLI for starting the daemon?

ketiltrout:
I hadn't thought about changing instances of "service" but it's worth looking at.  I don't mind using "service" for the daemon, but we should probably use the same word everywhere, if we can.  Especially in the config file.

`alpenhorn/doc` and `alpenhorn/demo` are very out-of-date.  I'm not going to touch them until the client is stable.

Similarly, `alpenhorn.cli.transport` is still unfixed.  It's not clear what's going to happen to it, but I'm going to leave it alone for now.

>I presume the files tests/cli/test_client*.py don't matter since they seem be skipped.

Yeah, those are old tests which haven't been updated yet.  They'll be renamed or deleted as the CLI fixing proceeds.

>Finally, I found the function deamon.service.cli with the docstring Alpenhorn data management service. This got me thoroughly confused :) Is this a separate CLI for starting the daemon?

Yes, this is the program-entry point for the daemon, which does make it a command-line interface, but it's probably best to rename it to something else to avoid confusion.

ketiltrout:
That's for the look-over.  I've replace most uses of "service" now with "daemon", too, so we're using the same term everywhere.

rikvl (approved):

PR #221, opened 2024-11-11
fix(client): First part of "node" command group fix

### LINKED ISSUE #39: `create_node` should include the option for setting StorageNode.username

ketiltrout:
This fixes the simpler commands in the "node" group:
* `activate`: activate a node
* `create`: create a node.  This closes #39
* `deactivate`: deactivate a node
* `list`: list nodes with metadata
* `rename`: rename a  node
* `stats`: show file counts and sizes for a node.  Replaces the old top-level `status` command

The `node active` command has been removed.  It's now handled by `list`:
```
  alpenhorn node list --active
```

Left to do are `clean` and `import` (AKA `scan` or `import-files`)

## Other changes

The CLI entry point has been moved to `client.entry` to allow `client.cli` to be a general sort of dumping ground for common CLI functions.

I wrote a custom `assert_row_present` test to check the output rows in the `tabulate` tables that the CLI produces.  Not only does it save repeated use of awkward regexes, it replaced the cryptic (and tautological) "None is not None" error message for a failed match with something that indicates the row that was being looked for.  Have updated some of the previous CLI tests to use this.

In the interests reducing verbosity.  I changed "Needs Check" to "Suspect" (which has always been a term we use for `has_file='X'` files, so I'm happy with that) and "Pending Removal" with "Released" for files about to be deleted (which I'm less sold on, but there shouldn't be many of these around, if the daemon is running, so maybe it's fine.)


ketiltrout:
Oh, looks like I forgot `node show` here.  Let me add that ...

ljgray (approved):

ketiltrout:
Gah. I forgot about `node show` again.  I guess I'll make another PR.

PR #224, opened 2024-11-13
feat(cli): Add "node show" command

ketiltrout:
Also implement the `--node-stats` option to "group show"

Should the stat-generating function `get_stats` stay in `client.node.stats` or should it be moved to `client.cli` (which contains general-purpose CLI library functions)?


ljgray (approved):
Looks fine. I'm for leaving `get_stats` where it is

PR #226, opened 2024-11-14
feat(daemon): Permit deletion of bad files

### LINKED ISSUE #225: Daemon should be able to delete suspect/corrupt files
i.e.  we should attempt to delete `has_file='X'/'M', wants_file='N'` copies.

https://github.com/radiocosmology/alpenhorn/blob/56348b109b35ceeaa62b8ca3dd12eaccaaf4e903/alpenhorn/daemon/update.py#L392

* Corollary: daemon shouldn't attempt to check files with `has_file='M', wants_file='N'`, because we're just going to delete them.
* Also need a client option to be able to flag them: `node clean --include-bad`

ketiltrout:
Updates the daemon to try deleting any file with `has_file!='N'` (instead of just the `='Y'` ones).

The daemon will also skip trying to check files released for immediate deletion (`has_file=='N'`).

Also added `--include-bad` to CLI `node clean` command to permit manipulating `has_file` for suspect and corrupt files.

Closes #225

### DIFF
--- a/alpenhorn/cli/node/clean.py
+++ b/alpenhorn/cli/node/clean.py
@@ -1,5 +1,7 @@
 """alpenhorn node clean"""
 
+from __future__ import annotations
+
 import click
 import datetime
 import peewee as pw
@@ -25,6 +27,7 @@ def _run_query(
     acq,
     archive_ok,
     days: datetime.datetime,
+    has_file: pw.Expression,
     size: int,
     target,
     clean_goal: str,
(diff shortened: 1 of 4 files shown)


alpenhorn/cli/node/clean.py
```diff
@@ -51,6 +52,8 @@ def _run_query(
         Replaces the `now` and `cancel` parameters.
     days:
         Converted to datetime
+    has_file:
+        peewee constraint for ArchiveFileCopy.has_file
```
ljgray, line 56: Minor, but it would be nice to show an example of what would be passed here (i.e., Example: ArchiveFileCopy.has_file == "Y")

ljgray (approved):

PR #228, opened 2024-11-14
feat(CLI): StorageTransferAction support

ketiltrout:
This is based on my suggestion to remove the `action` group in #202, which I think was ill-defined.

Instead:

* Adds an `--actions` flag to `node show` and `group show` to show relevant `StorageTransferAction`s affecting the node/group.
* Adds an `--all` flag to `node show` and `group show` to provide a way to list everything without having to know what lists what.
* Adds two new commands to manipulate `StorageTransferAction`s:
  * `group auto-sync GROUP NODE` to set/clear the autosync flag
  * `node auto-clean NODE GROUP` to set/clear the autoclean flag

For a given (NODE, GROUP) pair, these two commands modify the same row in the `StorageTransferAction` table (the pair has to be unique in the table).  So, there's ambiguity over which command goes with which "noun". The way I've done it, the command is associated with the object that the auto-action modifies (auto-sync adds files to GROUP; auto-clean deletes files from NODE).  There's probably an argument for doing it the other way; I'm not sure it matters much, so long as we are consistent about it.

Also: should the command names have a dash in them?  Or would `group autosync GROUP NODE` etc. be better?

### DIFF
--- a/alpenhorn/cli/group/__init__.py
+++ b/alpenhorn/cli/group/__init__.py
@@ -5,6 +5,7 @@ import peewee as pw
 
 from ...db import StorageGroup, StorageNode
 
+from .autosync import autosync
 from .create import create
 from .list import list_
 from .modify import modify
@@ -17,6 +18,7 @@ def cli():
     """Manage Storage Groups."""
 
 
+cli.add_command(autosync, "autosync")
 cli.add_command(create, "create")
 cli.add_command(list_, "list")
 cli.add_command(modify, "modify")
--- a/alpenhorn/cli/node/__init__.py
+++ b/alpenhorn/cli/node/__init__.py
@@ -13,6 +13,7 @@ from ...common import util
 from ...db import ArchiveAcq, ArchiveFile, ArchiveFileCopy, StorageGroup, StorageNode
 
 from .activate import activate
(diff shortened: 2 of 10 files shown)


alpenhorn/cli/group/__init__.py
```diff
@@ -17,6 +18,7 @@ def cli():
     """Manage Storage Groups."""
 
 
+cli.add_command(autosync, "auto-sync")
```
ljgray, line 21: In response to your question - I personally prefer no dash or underscore (so `autosync` here) because I end up forgetting whether it's a dash or an underscore

alpenhorn/cli/node/__init__.py
```diff
@@ -31,6 +32,7 @@ def cli():
 
 
 cli.add_command(activate, "activate")
+cli.add_command(autoclean, "auto-clean")
```
ljgray, line 35: Same here

ljgray (approved):

alpenhorn/cli/group/__init__.py
ketiltrout: That's reasonable.  I can change it.

Issue #141, opened 2022-11-16
The `verify` command assumes the target node is locally available.


ketiltrout:
The client needs to check whether the node is available locally or not at the start.  Then, if the node is non-local, it should explain to the user that the `verify` command needs to be run on the host that node is present on.

Right now, the client will tell the user that the node is corrupt when the node is not local (because it will fail to read the obviously missing `ALPENHORN_NODE` file), which isn't helpful.

ketiltrout:
With #229, the `verify` command no longer does any file checking in the CLI, so this is no longer a problem.

Issue #133, opened 2020-10-13
Can't re-check file copies marked "has_file=M"


cubranic:
I think I ran into a loophole where file copies are marked "M" and there is no way to restore them short of hand-editing the database.

This is what happened:

1. transport disk at UBC, all files on it are good, and marked as "has_file=Y" in the database
2. copy request from the TD to cedar
3. on cedar, alpenhornd starts pulling the files, but I run out of quota space, and bbcp returns an error as it tries to write them.
4. alpenhorn detects error during copying, and marks the file copy on the TD as "has_file=M".
5. there are no ways to reset that to "Y" using client utilities, and I have to write SQL in order to do so and resume the copying after fixing my quota issues.

Commands I tried:
- `verify`: only checks files which are "has_file=Y" to mark them "N" or "X" if they fail
- `scan`: detects the file copy is already registered and does not make any checks or changes to the status

ketiltrout:
This happens all the time with our UBC-to-cedar copying, too.  At least with alpenhorn 1, the solution is to run an alpenhorn daemon on the machine at UBC with the TDs (`jingle` in our case).  The _only_ thing that particular daemon _ever_ does is wait around for the `cedar` daemon to mark TD files as `has_file='M'` and then runs a verify on them to either mark them `X` or `N`.

Other that that (i.e. the vast, vast majority of the time), it just spins there doing nothing.

That said, a manual `verify` should also be checking `has_file="M"` entries.

ketiltrout:
Really, verify should check all file copies that don't have `has_file='N'`, i.e. all supposedly extant, files whether they're ok (`Y`), uncertain (`M`) or known to be bad already (`X`), and then reporting differences from the database.

jrs65:
Yeah, I agree the condition should probably `has_file!="N"` for `verify`

ketiltrout:
Then it's complementary to `scan`, which is meant to find files which aren't expected to be on that node (i.e. it only updates absent entries and entries with `has_file=N`).

PR #229, opened 2024-11-15
fix(CLI): Re-implement "node verify"

ketiltrout:
This is a re-implementation of "node verify" which removes all I/O from the CLI.

As previously discussed, I think it's difficult to have I/O in the client.  In general, it's difficult to keep I/O clean unless we're
very specific on how I/O works, which we've become less-so with the new I/O framework stuff.  So, keeping I/O to the daemon vastly simplifies things.

Even with the preexisting simplified version of verification that I'm replacing, issues like #141 show how easy it is to get in to situations where verification fails.  The verify-in-cli scheme requires both running where the node is hosted and from a user that has access to the files on the node.

I don't know how we would ever run a CLI-based verify on a node like `cedar_nearline` where it would have to wait for files to be recalled.

This new verify command can also be used to "cancel" a verification, which really means force the `has_file='M'` suspect file into one of the other `has_file` values.

I've taken the check-confirm-update stuff from my implementation of `clean` and put it in a utility function so it can be used elsewhere (like here).

With this PR, I think neither of the issues linked to the verify command are relevant anymore:

* With the removal of I/O from the CLI, where its run is now immaterial: #141
* The underlying problem with #133 was trying to find a way to check files on Storage Nodes hosted where there was no daemon running.  In alpenhorn1 you could do some of the daemon's work in the client, but it was never a complete solution, and in practice we always ended up putting daemons on machines where we had nodes (e.g. the transport disk off-load machine here at UBC), even if that daemon spent most of its time doing nothing.

The current answer to #133 would be: run the daemon to maintain the node.

There are also a couple of commits in this PR tidying up minor things I overlooked in the previous PRs.

### DIFF
--- a/alpenhorn/cli/cli.py
+++ b/alpenhorn/cli/cli.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import click
 import datetime
 import peewee as pw
+from typing import TYPE_CHECKING
 
 from ..common.logger import echo as echo
 from ..db import (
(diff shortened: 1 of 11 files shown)


ketiltrout:
>  In alpenhorn1 you could do some of the daemon's work in the client, but it was never a complete solution

Actually, maybe we should have an option for the daemon which tells it to run once, wait for I/O to finish and then exit.  Then people could run the daemon as an on-demand foreground process, in these sorts of cases, rather than forcing it to run as a long-running background process.

PR #235, opened 2024-11-18
feat(daemon): Import request functionality

ketiltrout:
This adds a new table to the data index: ArchiveFileImportRequest, which contains requests for the daemon to import new file(s) into the data index.

Each request contains a node in which the import is done, a path (relative to node.root) to import, and two flags:

* recurse: whether path should be recursively scanned for files to import
* register: whether new files should be added to the data index, or whether the import should just add new ArchiveFileCopy records for pre-registered ArchiveFiles.

A third `completed` flag indicates to the daemon whether the request has been handled.

In terms of the update loop itself, there's not a whole lot to add. The `auto_import` module already has all the necessary functionality to handle these requests, and I've added a new function (`update_import`) to the `UpdateableNode` class in `daemon.update` to find all pending import requests for a node and dispatch them to the appropriate `auto_import` function.

There are two such functions in the `auto_import` module:

* `import_file` can be used to import a single file (i.e. when the `recurse` flag is False.  This is the same function that is called by the auto-import watchdogs to import newly discovered files.
* for the `recurse == True` case, `scan`, which has been renamed from `auto_import.catchup` is a Task-run function to recursively scan the given path and call `import_file` on unimported files found.

The functionality of the two functions haven't seen a whole lot of change from what was in `auto_import` previously.  `scan` now takes a path relative to `node.root` (which can be `"."` to scan the whole node).  Both functions get passed the `register` flag from the import request, and functionality has been adeed to the inner `_import_file` function to handle the case when we should import known files but not register new ones.

Finally, both functions take the `ArchiveFileImportRequest` record itself so they can mark it as completed once the operation finishes.

The CLI-facing side to this functionality will be added in a second PR.

### DIFF
--- a/alpenhorn/db/archive.py
+++ b/alpenhorn/db/archive.py
@@ -85,7 +85,7 @@ class ArchiveFileCopy(base_model):
 
 
 class ArchiveFileCopyRequest(base_model):
-    """Requests for file copies.
+    """Requests for file transfer from node to group.
 
     Attributes
     ----------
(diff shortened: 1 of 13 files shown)


alpenhorn/db/archive.py
```diff
@@ -118,3 +118,40 @@ class ArchiveFileCopyRequest(base_model):
...
+    register = pw.BooleanField(default=False)
+    completed = pw.BooleanField(default=False)
+    timestamp = pw.DateTimeField(default=pw.utcnow, null=True)
+
+    def complete(self) -> None:
+        """Set this request to complete in the database."""
+
+        # If we're already complete, this function does nothing
+        if not self.completed:
+            self.completed = 1
```
ljgray, line 156: why `1` instead of `True`?

ljgray (approved):

alpenhorn/db/archive.py
ketiltrout: It's the same, though maybe it would be clearer to use True.  SQL doesn't have a boolean type, so when we make bool fields what we're actually making is a 1-byte int.  Peewee converts True/False to 1/0 when writing to the database.

I've been preferring 0/1 instead over False/True recently because of flake8 always complaining about the, e.g., `ArchiveFileCopyRequest == False` constructs in `where` clauses.  (Flake8 wants `is False` in this case, but that doesn't work for peewee because `is` can't be overloaded like `==` can.)

Maybe for assignment it would be better to use `True`, especially in this function where we see the `BooleanField` just above it.

ljgray: Makes sense. I'm not bothered either way, so I think it's fine to just stick with `1`

ketiltrout: I've changed it to `True`.  It's probably less confusing this way.

PR #237, opened 2024-11-18
feat(CLI): node init command

### LINKED ISSUE #227: Need a "node init" in client
We need some way to trigger "initialisation" of a node in the CLI (i.e. creation of the `ALPENHORN_NODE` file for a newly-created StorageNode.

ketiltrout:
Also a --init flag to "node create" to do the same thing.

This requests the daemon intialise the node, which typically means creating the ALPENHORN_NODE file, but doesn't have to.  The request is made via the ArchiveFileImportRequest with path="ALPENHORN_NODE" (which is otherwise a forbidden path to try to import).

This has involved some changes to the daemon to give it the ability to create ALPENHORN_NODE files.  The request can't be handled in the normal part of the loop (which is skipped for uninitialised nodes), so instead it's handled during the `check_init` test in the early part of the update loop.

I've also taken the opportunity to disambiguate two different uses of the term "active" which used to be in the daemon:

* the `StorageNode.active` bit is set
* the I/O layer is ready for I/O

The second of these is now called "initialised" and "active" now only refers to the first one.

Closes #227

### DIFF
--- a/alpenhorn/db/archive.py
+++ b/alpenhorn/db/archive.py
@@ -128,8 +128,8 @@ class ArchiveFileImportRequest(base_model):
     node : foreign key
         The StorageNode on which the import should happen
     path : string
-        The path to import.  If this is a directory and
-        recurse is True, it will be recursively scanned.
+        The path to import.  If this is a directory and recurse is True,
+        it will be recursively scanned.
     recurse : bool
         If True, recursively scan path for files, if path is a directory.
     register : bool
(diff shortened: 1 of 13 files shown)


alpenhorn/db/archive.py
```diff
@@ -139,6 +139,15 @@ class ArchiveFileImportRequest(base_model):
         Set to true when the import request has completed.
     timestamp : datetime
         The UTC time when the request was made.
+
+    If `path` is the special value "ALPENHORN_NODE", then the request is
+    a node initialisation request, instead of a normal import request.
+    This only happens on active nodes which aren't already initialised: a
+    initialisation request on an already initialised not is ignored.
```
ljgray, line 146:
```suggestion
    initialisation request on an already initialised node is ignored.
```

ljgray (approved):

PR #242, opened 2024-11-19
feat(util): Make `pretty_bytes` more forgiving

ketiltrout:
While it was pretty easy to avoid passing weird stuff to `pretty_bytes` in ths daemon, that's harder to do in the CLI.

I've found I've been doing the same None check a lot before using the function to deal with the potential for nulls in the database.

So, let's make the function a little more forgiving about the input. Not sure if "??? B" is the best thing to return in the None case. Let me know if you have a preference for something else.

### DIFF
--- a/alpenhorn/common/util.py
+++ b/alpenhorn/common/util.py
@@ -220,19 +220,19 @@ def get_hostname() -> str:
     return socket.gethostname().split(".")[0]
 
 
-def pretty_bytes(num: int) -> str:
+def pretty_bytes(num: int | None) -> str:
     """Return a nicely formatted string describing a size in bytes.
 
     Parameters
     ----------
-    num : int
+    num : int or None
         Number of bytes
 
     Returns
     -------
     pretty_bytes : str
-        A formatted string using power-of-two prefixes,
-        e.g. "103.4 GiB"
+        If `num` was None, this will be "-".  Otherwise, it's a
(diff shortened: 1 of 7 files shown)


alpenhorn/common/util.py
```diff
@@ -242,8 +242,13 @@ def pretty_bytes(num: int) -> str:
         `num` was less than zero
     """
 
+    # Return something unhelpful if given None
+    if num is None:
+        return """??? B"""
```
ljgray, line 247: I don't think this needs triple quotes

ljgray (approved):
Looks fine. I'm not sure that I love "??? B" as a stand-in for `None` because seeing something return/print that would make me think that something's gone wrong. I think that `"-"` would be fine like you were using before, or something else more neutral.

alpenhorn/common/util.py
ketiltrout: Dah.

ketiltrout:
> I think that "-" would be fine like you were using before, or something else more neutral.

Sure, we can do that.

I originally thought "??? B" was better because it includes the unit "B", so we couldn't get into situations where you have a "-" with no context, say:
```
   Cleaning 17 files (-).
```
but I realise now I'm dealing with that situation anyways because I want it to say `17 files (unknown size)` and not `17 files (??? B)`, or whatever.

PR #247, opened 2024-11-22
feat(CLI): "file list" command

### LINKED ISSUE #45: Command to show known corrupt files (was: `verify` should show files already known to be corrupt)
The query to get the file list in `client.verify` has a condition "`ar.ArchiveFileCopy.has_file == 'Y')`", so that known corrupt files will not be shown in the output. This is kind of confusing to the user, as there is no insight into corrupt files.

Furthermore, even the files that were only discovered to be corrupt during the `verify`'s run, and so their `has_file` attribute gets changed from 'Y' to 'X', will disappear from the output if `verify` was run again immediately.

ketiltrout:
Usage here is a little complex.  Might need better help to explain which options can be used together.  Also: is `--not` a too confusing for the flag which negates the node/group constraint?

I found it difficult to collapse this command into a single query, so instead there are three stages (excluding the various name-to-thing look-ups):

1. one or more query to handle location and state flag constraints
2. a query, maybe, for --from and --to
3. a final query that ties the above together with the --acq limit, and applies the --not, if necessary.

Closes #45 via:
```
file list --corrupt --node=<NODE>
```

### DIFF
--- a/alpenhorn/cli/file/list.py
+++ b/alpenhorn/cli/file/list.py
@@ -116,14 +116,13 @@ def list_(
     Limit by location
     -----------------
 
-    To limit the list based on where files are or are not present, use
-    the --absent-node=NODE, --absent-group=GROUP, --node=NODE and/or --group=GROUP
+    To limit the list based on where a file is, or is not, present, use the
+    --absent-node=NODE, --absent-group=GROUP, --node=NODE and/or --group=GROUP
     options to list places to search.
 
-    The first two of these negative constraints (nodes and/or groups which must not
-    have the files) and the second two are positive constraints (nodes and/or groups
-    which must have the file).  For a file to be listed, it must satisfy both the
-    negative and positive constraints.
+    The first two of these options are negative constraints (indicating nodes or
+    groups which must not have the file) and the last two are positive constraints
+    (nodes or groups which must have the file).
 
     Normally, only one of the location constraints given need be satisfied for a
     file to be listed, but if you also use "--all", then a file will be listed only
@@ -203,10 +202,10 @@ def list_(
     # from DeMorgan's rule), and "not all_", as expected, for the positive ones.
     if absent_node:
         absent_nodes = resolve_node(absent_node)
-        absent_node_files = files_in_nodes(absent_nodes, state_expr, in_any=all_)
+        absent_node_files = files_in_nodes(absent_nodes, None, in_any=all_)
     if absent_group:
         absent_groups = resolve_group(absent_group)
-        absent_group_files = files_in_groups(absent_groups, state_expr, in_any=all_)
+        absent_group_files = files_in_groups(absent_groups, None, in_any=all_)
     if node:
         nodes = resolve_node(node)
         node_files = files_in_nodes(nodes, state_expr, in_any=not all_)
@@ -215,7 +214,7 @@ def list_(
         group_files = files_in_groups(groups, state_expr, in_any=not all_)
 
     # In --details mode, extra node or group details are provided in very
-    # restricted circumstatnces:
+    # restricted circumstances:
     #  1. exactly one --group or exactly one --node was specified
     #  2. not restricted by syncability (i.e. no --from or --to)
     detail_node = None
--- a/alpenhorn/cli/file/list.py
+++ b/alpenhorn/cli/file/list.py
@@ -1,0 +1,342 @@
+"""alpenhorn file list command."""
+
+import click
+import peewee as pw
+from tabulate import tabulate
+
+from ...common.util import pretty_bytes
+from ...db import (
+    ArchiveAcq,
+    ArchiveFile,
+    ArchiveFileCopy,
+    ArchiveFileCopyRequest,
+    StorageGroup,
+    StorageNode,
+)
+from ..cli import echo, pretty_time
+from ..options import (
+    cli_option,
+    not_both,
+    both_or_neither,
+    files_in_nodes,
+    files_in_groups,
+    resolve_acqs,
+    resolve_group,
+    resolve_node,
+    state_constraint,
+)
+
+
+def _state_flag_help(name):
+    """Returns help for one of the state flags"""
+
+    return "Limit to " + name + "files.  Must be accompanied by at least one "
+    "--node or --group to provide a place to probe file state.",
+
+
+@click.command()
+@cli_option("acq")
+@cli_option("all_", help="Limit to files that exist on all NODEs and GROUPs provided.")
+@click.option("--corrupt", is_flag=True, help=_state_flag_help("corrupt"))
+@click.option("--details", is_flag=True, help="Show details for listed files.")
+@click.option(
+    "from_",
+    "--from",
+    metavar="SRC",
+    help="Must be specified with --to=DEST: limit to files that can be synced "
+    "from Node SRC to Group DEST",
+)
+@click.option(
+    "--group",
+    metavar="GROUP",
+    multiple=True,
+    help="Limit to files present (or not present) in GROUP.  "
+    "May be specified multiple times.",
+)
+@click.option("--healthy", is_flag=True, help=_state_flag_help("healthy"))
+@click.option("--missing", is_flag=True, help=_state_flag_help("missing"))
+@click.option(
+    "--node",
+    metavar="NODE",
+    multiple=True,
+    help="Limit to files present (or not present) on NODE.  "
+    "May be specified multiple times.",
+)
+@click.option(
+    "not_", "--not", is_flag=True, help="Limit to files absent from NODEs and GROUPs"
+)
+@click.option("--suspect", is_flag=True, help=_state_flag_help("suspect"))
+@click.option(
+    "--to",
+    metavar="DEST",
+    help="Must be specified with --from=SRC: limit to files that can be synced "
+    "from Node SRC to Group DEST",
+)
+@click.pass_context
+def list_(
+    ctx,
+    acq,
+    all_,
+    corrupt,
+    details,
+    from_,
+    group,
+    healthy,
+    missing,
+    node,
+    not_,
+    suspect,
+    to,
+):
+    """List Files.
+
+    Without options, lists all Files registered to the Data Index.
+    There are several ways to limit the list of files returned:
+
+    \b
+    Limit by acqusition
+    -------------------
+
+    It is always possible to limit the list to only files in acqusition(s)
+    specified in the --acq option.
+
+    \b
+    Limit by location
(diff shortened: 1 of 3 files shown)


rikvl:
> `file list --not --corrupt --node=Node`
lists all files which don't have a corrupt copy on node Node (i.e., as the usage says, --not applies to the node selection only).

Wouldn't this example cause a `UsageError` because of the following?
https://github.com/radiocosmology/alpenhorn/blob/38e87f623db174d47fef8fc24106af093f35a9c1/alpenhorn/cli/file/list.py#L172-L173

tests/cli/file/test_list.py
```diff
@@ -0,0 +1,649 @@
...
+
+    group = StorageGroup.create(name="Group")
+    StorageNode.create(name="Node", group=group)
+
+    cli(2, ["file", "list", "--from=Node"])
+    cli(2, ["file", "list", "--to=Group"])
+
+
+def test_state_not(clidb, cli):
+    """state flags  and --not can't be used together."""
```
rikvl, line 26:
```suggestion
    """State flags and --not can't be used together."""
```

tests/cli/file/test_list.py
```diff
@@ -0,0 +1,649 @@
...
+
+    group = StorageGroup.create(name="Group")
+    StorageNode.create(name="Node", group=group)
+
+    for state_flag in ["--corrupt", "--missing", "--healthy", "--suspect"]:
+        cli(2, ["file", "list", state_flag, "--not", "--node=Node"])
+
+
+def test_state_to_from(clidb, cli):
+    """state flags can't be used with --to or --from."""
```
rikvl, line 36:
```suggestion
    """State flags can't be used with --to or --from."""
```

tests/cli/file/test_list.py
```diff
@@ -0,0 +1,649 @@
...
+
+
+def test_bad_group(clidb, cli):
+    """Test a non-existent --group"""
+
+    cli(1, ["file", "list", "--group=MISSING"])
+
+
+def test_bad_group(clidb, cli):
+    """Test a non-existent --group"""
```
rikvl, line 75: I think this should be different from the above

tests/cli/file/test_list.py
```diff
@@ -0,0 +1,649 @@
...
+
+    StorageGroup.create(name="Group")
+
+    cli(1, ["file", "list", "--to=Group", "--from=MISSING"])
+
+
+def test_bad_group(clidb, cli):
+    """Test a non-existent --group"""
```
rikvl, line 83: Idem

rikvl (approved):
Found a few typos in comments and docs

rikvl:
> is --not a too confusing for the flag which negates the node/group constraint?

At first I was going to suggest perhaps using something more verbose, like `--negate-location-contraint`, but now I think it's good the way it is.

rikvl:
> Also, the distinction between --not and --not --all might be a little subtle:
   `file list --not --node=Node1 --node=Node2`
lists files that are absent from both Node1 and Node2
   `file list --not --all --node=Node1 --node=Node2`
lists files which are absent from at least one of Node1 or Node2. Does that feel backwards?

It's clearly explained in the documentation. And changing this to the opposite could end up more confusing

ketiltrout:
> Wouldn't this example cause a `UsageError` because of the following?

Ah, right, it would.  I think I ended up prohibiting this construct due to it not making a whole lot of sense.

ketiltrout:
After sleeping on it, what do you think about this:

Instead of a `-not` flag, just have more options to explicitly look for files not on thing: `--absent-from-node=NODE` and `--absent-from-group=GROUP` with maybe shorter names.

Might both remove ambiguity and make the command more flexible because you can mix absence and presence.  In fact, in with these, the syncability search `--from=NODE --to=GROUP` is essentially just `--node=NODE --absent-from-group=GROUP` (though I don't think I'd get rid of the special-case `--from/--to`)

alpenhorn/cli/file/list.py
```diff
@@ -0,0 +1,342 @@
...
+
+    Using both "--from=SRC" and "--to=DEST" will limit the list to files which
+    can be synced (copied) from the Node SRC to the Group DEST.  Specifically,
+    this restricts the list to files which are healthy on SRC and absent from
+    DEST.
+
+    You cannot simultaneously limit both by file state and syncability.
+    """
+
+    # Usage checks.
```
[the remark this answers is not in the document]
ketiltrout: Yes, `--details` is always allowed.  It just changes how the output is present.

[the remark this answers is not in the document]
ketiltrout: The "extra details" mode is limited, but you can always use `--details` to get the `ArchiveFile` contents tabulated.

alpenhorn/cli/file/list.py
```diff
@@ -0,0 +1,342 @@
...
+            )
+
+    # Figure out the state constraint from the flags
+    state_expr = state_constraint(
+        corrupt=corrupt, healthy=healthy, missing=missing, suspect=suspect
+    )
+
+    # Do we want the intersection or union of the nodes and groups?
+    # This is just XOR of --all and --not
+    intersect = all_ ^ not_
```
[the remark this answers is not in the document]
ketiltrout: I think the subtlety of why the XOR of `--not` and `--all` determines whether we use union or intersection is the same difficulty that the user might have when they try to figure out if they need `--not` or `--not --all`.  So, this might be another argument to drop `--not` and replace it with `--absent-node` &c. as I suggest below.

ketiltrout:
NOTE: I've pulled the CLI common-code refactoring out of this PR and made a new one for it over here: https://github.com/radiocosmology/alpenhorn/pull/250 This is now just the command implementation.

rikvl:
> After sleeping on it, what do you think about this:
> 
> Instead of a `-not` flag, just have more options to explicitly look for files not on thing: `--absent-from-node=NODE` and `--absent-from-group=GROUP` with maybe shorter names.
> 
> Might both remove ambiguity and make the command more flexible because you can mix absence and presence. In fact, in with these, the syncability search `--from=NODE --to=GROUP` is essentially just `--node=NODE --absent-from-group=GROUP` (though I don't think I'd get rid of the special-case `--from/--to`)

What's currently non-intuitive is that `--not` only applies to location constraints and not to, say, file state constraints. These new options would make that very clear

ketiltrout:
Yeah, that's what I'm thinking.  `--not` is a very vague term for a flag.

ketiltrout:
OK, I've removed `--not` and in its place added `--absent-node` and `--absent-group` for explicit selection of a negative location constraint.

I think these are the results:
* Improved the flexibility of the command, which is good, since you can now mix positive and negative constraints instead of having to choose one or the other (though there's still only one `--all` for both types)
* I think the code is clearer on how the selections are made and combined, which is also good
* Might be more intuitive for users, but I suspect negative constraints are going to cause confusion however they work.

Given the descriptions of the options, and the realisation that a file is in a group if it is on _at least one_ node in the group, and after some mental Venn diagramming, it's likely a user will figure out why, if StorageGroup `Group` contains `Node` and `Node2`, the positive constraint `--group=Group` is the same as `--node=Node1 --node=Node2`, but you need to add `--all` to convert the negative constraint `--absent-group=Group` into an equivalent node-based one, i.e. `--all --absent-node=Node1 --absent-node=Node2`, but I don't know if I'd call it intuitive.

Is it an improvement?

rikvl:
I.e., while in general there are ORs between individual constraints (which can be turned into an AND by `--all`), there's always an implicit AND between the positive constraints and the negative constraints.

ketiltrout:
OK, `--all` now also applies to how positive and negative constraints are combined.  This has made the `--from --to` files a bit special, because that is always AND-ed with any other location constraint (ignoring whether `--all` was used).

alpenhorn/cli/file/list.py
```diff
@@ -0,0 +1,382 @@
...
+        absent_group_files = files_in_groups(absent_groups, state_expr, in_any=all_)
+    if node:
+        nodes = resolve_node(node)
+        node_files = files_in_nodes(nodes, state_expr, in_any=not all_)
+    if group:
+        groups = resolve_group(group)
+        group_files = files_in_groups(groups, state_expr, in_any=not all_)
+
+    # In --details mode, extra node or group details are provided in very
+    # restricted circumstatnces:
```
rikvl, line 218:
```suggestion
    # restricted circumstances:
```

alpenhorn/cli/file/list.py
```diff
@@ -0,0 +1,382 @@
...
+    Limit by location
+    -----------------
+
+    To limit the list based on where files are or are not present, use
+    the --absent-node=NODE, --absent-group=GROUP, --node=NODE and/or --group=GROUP
+    options to list places to search.
+
+    The first two of these negative constraints (nodes and/or groups which must not
+    have the files) and the second two are positive constraints (nodes and/or groups
+    which must have the file).  For a file to be listed, it must satisfy both the
```
rikvl, line 125:
```suggestion
    The first two of these are negative constraints (nodes and/or groups which must not
    have the files) and the second two are positive constraints (nodes and/or groups
    which must have the files).  For a file to be listed, it must satisfy both the
```

rikvl (approved):
Thanks for all this work, Don! I think the location constraints are now quite intuitive

alpenhorn/cli/file/list.py
```diff
@@ -0,0 +1,382 @@
...
+    negative and positive constraints.
```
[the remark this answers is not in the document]
ketiltrout: Yeah, it definitely needs to be changed.  Thanks for catching that.

alpenhorn/cli/file/list.py
```diff
@@ -0,0 +1,382 @@
...
+    # (checking --from is enough, given the previous test)
+    not_both(from_, "from", state_flag, state_flag)
+
+    # Can't use --all if we haven't used a location constraint
+    if all_:
+        if not node and not group and not absent_group and not absent_node:
+            raise click.UsageError(
+                "--all cannot be used without a location constraint "
+                "(one or more of --absent-group, --absent-node, --group, --node)"
+            )
```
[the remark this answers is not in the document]
ketiltrout: Hmm, true.  Using `--all` with a single constraint isn't going to cause trouble, but you're right there needs to be more than one for it to have any effect.

Really, there's no problem with using `--all` with _no_ location constraint.  It would just be ignored.  The usage check here is mostly to re-inforce that `--all` applies to location constraints only.

I'm not sure it's worth counting how many location constraints have been used.

PR #251, opened 2024-11-22
feat(CLI): "file create" command

ketiltrout:
Not the way you should be creating ArchiveFiles, but provided for completeness (like "acq create" was).

Has some actual I/O in it: it can scan a file to generate a MD5 hash if you want it to.  Other than that, this is another simple command: it just creates an ArchiveFile record.


ketiltrout:
I've realised there's probably some lack of name vetting going on here, but I'm going to leave dealing with that for now: #252

PR #255, opened 2024-11-23
feat(CLI): "file modify" command

ketiltrout:
All this lets you do is modify the md5sum or size_b of an ArchiveFile record.

If a record is modified, by default, it will mark all the copies as needing check because we don't want to get into a situation where the has_file value is wrong.

### DIFF
--- a/alpenhorn/cli/file/modify.py
+++ b/alpenhorn/cli/file/modify.py
@@ -1,0 +1,82 @@
+"""alpenhorn file show command."""
+
+import click
+import peewee as pw
+
+from ...db import ArchiveFile, ArchiveFileCopy, database_proxy
+from ..cli import echo, update_or_remove
+from ..options import cli_option, file_from_path, validate_md5
+
+
+@click.command()
+@click.argument("path", metavar="FILE")
+@cli_option("md5")
+@click.option("--no-reverify", is_flag=True, help="Don't reverify File copies on Nodes")
+@cli_option("size")
+@click.pass_context
+def modify(ctx, path, md5, no_reverify, size):
+    """Change File details.
+
+    This command can be used to change the MD5 hash
+    and/or recorded size of the File given by the path FILE
+    in the data index.  FILE should be specified as
+    "acq-name/filename".
+
+    By default, after updating a File record, all extant
+    copies of the File will be marked as needing reverification.
+    This is to ensure that all copies of the File marked as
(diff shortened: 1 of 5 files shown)


alpenhorn/cli/file/modify.py
```diff
@@ -0,0 +1,82 @@
...
+
+        if not updates:
+            echo("No change.")
+            ctx.exit()
+
+        # Do the update
+        ArchiveFile.update(**updates).where(ArchiveFile.id == file_.id).execute()
+
+        # Re-verfiy if, not prohibited
+        if not no_reverify:
```
ljgray, line 64: What is the rationale behind this being optional? I would think that we would _always_ want this re-verification to happen.

ljgray (approved):
Looks fine, just one comment

alpenhorn/cli/file/modify.py
ketiltrout: If you have a daemon running, then, yes, you should never use this option.

But the idea behind an option like this in the CLI is that the user is running alpenhorn without using the daemon, so we should provide general-purpose tools for them.

Really, if they're using the daemon, they should never be using this command at all because the daemon should be managing their files.

Should I make the language stronger in the usage text?  Maybe something along the lines of:
```
@click.option("--no-reverify", is_flag=True,
   help="Don't re-verify File copies on Nodes.  This option should only be used in cases where you're not using "
   "the daemon to manage your Storage Nodes."
```

Even if we got rid of this exception, users would still be able to manually edit `has_file` via the yet-to-be-implemented `file state` command.

ljgray: Makes sense. I think stronger language in the usage text would be a good idea

ketiltrout: I don't know if it's possible to do in Click, but maybe we could have some sort of flag to enable dangerous commands.  i.e. something like:
```
alpenhorn --dange-mode file mode --no-reveify
```
to really separate the commands which should be rarely run in normal operation.

ljgray: I can't find anything within Click to do that. We could probably implement something directly, or we could just the built-in prompt option in Click to get people to confirm unsafe flags.

ketiltrout: Yeah, it doesn't seem likely, but I've explained in #258 a bit more of what I'd like to do.

PR #257, opened 2024-11-25
feat(CLI): "file clean" command

ketiltrout:
The first of three ArchiveFileCopy-focused commands in this group.

This one manipulates the `wants_file` column.  I've tried to make it similar to the more familiar "node clean" command.

It's quite straightforward.

### DIFF
--- a/alpenhorn/cli/file/clean.py
+++ b/alpenhorn/cli/file/clean.py
@@ -1,0 +1,107 @@
+"""alpenhorn file clean command."""
+
+import click
+import peewee as pw
+
+from ...db import ArchiveFileCopy, database_proxy
+from ..cli import echo
+from ..options import (
+    at_least_one,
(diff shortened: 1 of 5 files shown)


alpenhorn/cli/file/clean.py
```diff
@@ -0,0 +1,107 @@
...
+
+        count = query.execute()
+
+    if count:
+        if cancel:
+            if node:
+                echo(f'Cancelled cleaning for "{path}" on Node "{node.name}".')
+            else:
+                requests = "request" if count == 1 else "requests"
+                echo(f'Cancelled {count} cleaning requests for "{path}".')
```
ljgray, line 101:
```suggestion
                echo(f'Cancelled {count} cleaning {requests} for "{path}".')
```

ljgray (approved):

PR #259, opened 2024-11-27
feat(CLI): "file state" command.

ketiltrout:
This is the primary command for modifying ArchiveFileCopy records. It can also create them.

I think there's going to need to be a pass through the CLI to make all the file state descriptions uniform, but I'm going to leave that for now.


alpenhorn/cli/file/state.py
```diff
@@ -0,0 +1,234 @@
...
+            copy = None
+
+        # Set wants_file if the file is now present
+        if (not copy or copy.wants_file == "N") and now_present:
+            updates["wants_file"] = "Y"
+
+        if ready:
+            # We can only ready a file if it's present
+            if not copy or copy.has_file != "Y":
+                raise click.ClickException(f"can't set ready bit: file not present.")
```
ljgray, line 84:
```suggestion
                raise click.ClickException("can't set ready bit: file not present.")
```

ljgray (approved):

PR #260, opened 2024-11-28
fix(import): Vet acq and file names

### LINKED ISSUE #252: Vetting acq and file names
With the CLI commands "acq create" and "file create" taking arbitrary strings that end up being used for acq or file names, there's going to need to be some vetting of the names.

The daemon implicitly sets some limits on the names to make them sanitary, by extracting them from a path to a file on a Node, but the CLI doesn't have that, so it should at least prohibit:

* starting or ending with `/`
* two or more `/` in a row (`/////`)
* `/./` or `/../` path elements

Anything else?  We definitely need to prohibit acq and file names from escaping from `node.root`.  Though I'm more fine with names which simply produce unaccessible paths (since those are fairly harmless, outside of the acq/file they're associated with being unaccessible to the daemon).

If the combination `acqname/filename` formed a concrete path we could use standard path normalisation, but they don't (becasue the CLI can be run somewhere without any data present), so we need to do something else.

The daemon always forms a path like: `node_root/acq_name/file_name` so leading/trailing slashes can't force it out-of-tree, but internal `/../` might.

We _do_ need to protect against having both an acq named `"acq"` and one named, say, `"acq/"`, since those would end up referring to the same files on disk.

ketiltrout:
This creates a function, `common.util.invalid_import_path`, that takes an acq name, file name or both, and checks whether it can be used.

This is then used in four places to try to reduce the number of invalid acq/file names that make it into the database:

* in the "acq create" CLI command
* in the "file create" CLI command
* to check the acq-name returned by the import-detect extension
* when processing non-recursive ArchiveFileImportRequests

I have also added a check for recursive ArchiveFileImportRequests that the requested scan path is resolvable and in-tree.

I think it's still fairly easy to maliciously add such names to the database, but this should deal with inadvertently providing invalid names.

Closes #252

### DIFF
--- a/alpenhorn/common/util.py
+++ b/alpenhorn/common/util.py
@@ -313,3 +313,47 @@ def pretty_deltat(seconds: float) -> str:
 
     # For short durations, include tenths of a second
     return f"{seconds:.1f}s"
+
+
+def invalid_import_path(name: str) -> str | None:
+    """Is `name` invalid as an import path?
+
+    i.e., can `name` be used as an ArchiveAcq or
+    ArchiveFile name (or both, combined).
+
+    Returns
+    -------
+    rejection_reason: str or None
+        A string describing why `name` was rejected,
+        or None, if the name was valid.
+    """
+
+    # Can't be the null string
+    if name == "":
+        return "empty path"
+
+    # Can't simply be "." or ".."
+    if name == "." or name == "..":
+        return "invalid path"
+
+    # Can't start with "/" or "./" or "../"
+    if name.startswith("/") or name.startswith("./") or name.startswith("../"):
+        return "invalid start"
+
+    # Can't end with "/" or "/." or "/.."
+    if name.endswith("/") or name.endswith("/.") or name.endswith("/.."):
+        return "invalid end"
+
+    # Can't have multiple "/" in a row
+    if "//" in name:
+        return "repeated /"
+
+    # Can't have internal "/./"
+    if "/./" in name:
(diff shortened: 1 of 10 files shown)


ketiltrout:
I've relaised there's a problem with this that I need to fix...

ketiltrout:
OK, that wasn't as difficult as I thought it would be.  The issue in my original PR was there needs to be different checks for non-recursive and recursive `ArchiveFileImportRequest`s, which I have now implemented.

alpenhorn/common/util.py
```diff
@@ -313,3 +313,43 @@ def pretty_deltat(seconds: float) -> str:
...
+    if name.endswith("/") or name.endswith("/.") or name.endswith("/.."):
+        return "invalid end"
+
+    # Can't have multiple "/" in a row
+    if "//" in name:
+        return "repeated /"
+
+    # Can't have internal "/./" or "/../" elements
+    if "/./" in name or "/../" in name:
+        return "invalid path element"
```
ljgray, line 353: It might be helpful to include the invalid path element which is causing this, unless you want the outputs of this function to be static for dict lookups or something

ljgray (approved):

alpenhorn/common/util.py
ljgray: Fair enough

ketiltrout: I've done it.  I don't see why not.

PR #269, opened 2024-12-05
feat(daemon): Run-once mode

### LINKED ISSUE #230: Limited run-then-exit mode for the daemon
> Actually, maybe we should have an option for the daemon which tells it to run once, wait for I/O to finish and then exit.  Then people could run the daemon as an on-demand foreground process, in these sorts of cases, rather than forcing it to run as a long-running background process.

_Originally posted by @ketiltrout in https://github.com/radiocosmology/alpenhorn/issues/229#issuecomment-2477711216_

ketiltrout:
Adds an option -o/--once/--exit-after-update to the daemon to make it run the update loop once, wait for all tasks to finish (including deferred ones), and then exit cleanly.

The end-to-end test in `tests/daemon/test_entry.py` has been updated to use this mode (rather than what we were doing before which was using the `loop_once` fixture that co-opts the global abort).

Requires #268 to prevent the end-to-end test from taking ten minutes to complete.

Closes #230

### DIFF
--- a/alpenhorn/daemon/update.py
+++ b/alpenhorn/daemon/update.py
@@ -846,7 +846,9 @@ class UpdateableGroup(updateable_base):
             self.io.idle_update()
 
 
-def update_loop(queue: FairMultiFIFOQueue, pool: WorkerPool | EmptyPool) -> None:
+def update_loop(
+    queue: FairMultiFIFOQueue, pool: WorkerPool | EmptyPool, once: bool
+) -> None:
     """Main loop of alepnhornd.
 
     This is the main update loop for the alpenhorn daemon.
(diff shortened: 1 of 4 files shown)


alpenhorn/daemon/update.py
```diff
@@ -1017,12 +1022,33 @@ def update_loop(queue: FairMultiFIFOQueue, pool: WorkerPool | EmptyPool) -> None
...
-        remaining = config.config["daemon"]["update_interval"] - loop_time
-        if remaining > 0:
-            global_abort.wait(remaining)  # Stops waiting if a global abort is triggered
+        if once:
+            # If we're in Exit-after-update mode, wait for updates to complete
+            # and then return
+            first_time = True
+            while True:
+                if queue.qsize + queue.inprogress_size + queue.deferred_size == 0:
+                    log.info("Update complete.  Exiting.")
```
rikvl, line 1031:
```suggestion
                    log.info("Updates complete.  Exiting.")
```
For consistency with other comments / log messages (?)

alpenhorn/daemon/update.py
```diff
@@ -1017,12 +1022,33 @@ def update_loop(queue: FairMultiFIFOQueue, pool: WorkerPool | EmptyPool) -> None
...
+                    return
+
+                if first_time:
+                    first_time = False
+                    log.info("Waiting for updates to complete.")
```
rikvl, line 1036: Is the reason for this construction that you don't want the message "Waiting for updates to complete." in the logs if there were none?

rikvl (approved):
Looks good as far as I can tell

alpenhorn/daemon/update.py
ketiltrout: Yes.  Well, if it didn't wait.

Issue #253, opened 2024-11-22
Resuscitate the docker test thing?


ketiltrout:
There's a fairly robust test suite these days using `pyfakefs` and in-memory sqlite databases to provide a test arena, so docker isn't as necessary as it once was, but the docker image could still be handy to provide a space for interactive testing/development.

PR #272, opened 2025-01-23
fix(demo): Get the docker demo working again

ketiltrout:
This is a first pass at resuscitating and modernising the docker-based demo.

It's not completely done; some of the demo needs restructuring for modern alpenhorn, but basic functionality is restored, and it's working up to around step 8 of `demo-script.md`.

One thing that I think will need changing is removing the `demo/data` directory which the alpenhorn containers are mounting as a volume. Running the demo results in `root`-owned files in the directory, which isn't great.  I think it'll be possible to remove the `data` directory completely and use normal (i.e. docker-managed) volumes instead, which is likely what should be happening.

I've had to do some work on the `pattern_exporter.py` example extension, which is used by the demo.  There's a new init function to populate the tables for the demo, and a fix to the detection code.

The `demo/Dockerfile.alpenhorn` and `demo/alpenhorn.conf` files have been resurrected from the test suite, where they had been deleted in #265.

Partial fix for #253


ketiltrout:
I've tacked on a fix for an incorrect string that ruff happened to find.  I think it's probably too trivial for a separate PR.

ljgray (approved):
I don't see any issues

PR #276, opened 2025-02-27
ci: bump latest Python to 3.13

ketiltrout:
I don't see any blockers here.


ketiltrout:
Will need to update the branch protection rules here.

ljgray (approved):
I don't see any reason not to

PR #277, opened 2025-03-06
feat(daemon): Prometheus client

### LINKED ISSUE #166: Prometheus client for server?
Deploying alpenhorn2 to `wind` will require either:
* rewriting `alpenmonitor`, a Perl script which scrapes the alpenhorn log file to create prometheus metrics; or else:
* add a prometheus client into the alpenhorn server, which is probably a better idea, and long-term easier to maintain.

ketiltrout:
This adds a prometheus client to the daemon which will be started if the port it should listen on is specified in the alpenhorn config.

I've added metrics here and there.  I suspect some changes will be necessary to what data is produced, but that can happen iteratively. It's difficult to know how we actually want to instrument the process without some testing.  If you see any glaring omission, feel free to suggest it.  In addition to the explicit metrics I've defined, some basic stats about the python process itself, e.g. CPU load, memory usage, &c. are automatically produced by the prom client server.

Metrics themselves are represented by a new `Metrics` class which mostly wraps `prometheus_client.Gauge` and `prometheus_client.Counter`. I've tried to make a system where we should only have to define a given metric once, even if we want to access it from various places.  (The biggest annoyance with defining metrics is the need to specify the description, which is completely unneeded once the metric has been defined once, but is still required for each `Counter` or `Gauge` instance.)

For a very few metrics, I couldn't find a natural place in the existing codebase to define them, so they've been put in the `metrics.by_name` function from which they can be retrieved.

I tried, at least, to come up with a set of metrics that could replace the metrics created by the `alpenmonitor` that creates metrics for alpenhorn on wind.  But I looked at that, and there's shockingly few metrics that it's actually producing (it's mostly just the top three panels in the Alpenhorn grafana page).

Note: all metrics that the daemon presents are about the running daemon. It does not provide any of the metrics currently produced by the CHIME `archive_stats` job (which summarises the data index by directly inspecting the database). There's no plan to add them, but it would probably be a good idea to migrate that script over to this repo as a genericly useful script, or maybe as some sort of `alpenhorn db stats` command.

The `prometheus_client` dependency is optional.  If it can't be found at runtime, most of the metric-handling in `common.metrics` code does nothing.  The rest of alpenhorn is agnostic as to whether the `prometheus_client` module was available or not.

Closes #166

### DIFF
--- a/alpenhorn/db/archive.py
+++ b/alpenhorn/db/archive.py
@@ -162,11 +162,3 @@ class ArchiveFileImportRequest(base_model):
     register = pw.BooleanField(default=False)
     completed = pw.BooleanField(default=False)
     timestamp = pw.DateTimeField(default=pw.utcnow, null=True)
-
-    def complete(self) -> None:
-        """Set this request to complete in the database."""
-
-        # If we're already complete, this function does nothing
-        if not self.completed:
-            self.completed = True
-            self.save(only=[ArchiveFileImportRequest.completed])
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,33 +56,14 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
 
-    - name: Install pip dependencies
+    - name: Install with dependencies
       run: pip install -e .[test]
 
-    - name: Run tests
+    - name: Run tests without prometheus client
       run: pytest -v .
 
-  # run-docker-tests:
+    - name: Install prometheus client
+      run: pip install prometheus_client
 
-  #   runs-on: ubuntu-latest
-  #   steps:
-  #   - uses: actions/checkout@v2
-
-  #   - name: Set up Python 3.9
-  #     uses: actions/setup-python@v1
-  #     with:
-  #       python-version: 3.9
-
-  #   - name: Install pip dependencies
-  #     run: |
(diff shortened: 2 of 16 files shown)


alpenhorn/db/archive.py
```diff
@@ -162,11 +162,3 @@ class ArchiveFileImportRequest(base_model):
     register = pw.BooleanField(default=False)
     completed = pw.BooleanField(default=False)
     timestamp = pw.DateTimeField(default=pw.utcnow, null=True)
-
```
ketiltrout, line 165: This is now `import_request_done()` in `alpenhorn.daemon.auto_import`

ljgray (approved):
Looks good, but I think I like the idea of making it optional

ketiltrout:
Sure, I'll give that a try.

ketiltrout:
I've made `prometheus_client` optonal.  When not found at import time, the code in `common.metrics` mostly does nothing, without affecting calling into the module (meaning the rest of `alpenhorn` doesn't need to know whether or not the prom client is present).

I've updated the CI to run the unit tests both with and without `promtheus_client` installed.  When it's not installed, all the tests in`test_metrics.py` are skipped, but calling into `common.metrics` is still tested in the remainder of the test suite.

ljgray (approved):
Great!

PR #281, opened 2025-03-09
feat(demo): Update demo script

### LINKED ISSUE #253: Resuscitate the docker test thing?

ketiltrout:
This is a re-write of the script in `demo/demo-script.md`.  It hits most of the same feature points, but it fleshes out how to, step-by-step, run the demo and adds a large amount of explanation on what we're trying to demonstrate and why alpenhorn works the way it does.

I'd consider this a first draft.  Could use some polishing and probably more expansion in places.  Definitely needs some help with formatting.  Markdown might not be feature-full enough.  I may translate it into some other sort of document type.  No immediate plans for that.

But it's good enough that I'm happy to say this closes #253 .

The demo script from this PR can be read as rendered Markdown here: https://github.com/radiocosmology/alpenhorn/blob/demo_fixes/demo/demo-script.md

## Demo data deleted

On the functional side of things, I've deleted the `demo/data` directory which was being used to house the volumes in the demo containers.  This was a problem for a number of reasons, not the least being the demo would create a bunch of files there as root (or whatever root was mapped to by docker) as the demo proceeded.

Instead the demo now just uses normal docker-managed volumes, meaning they can easily be deleted if the user wants to clean up after the demo and/or reset the demo to the start (and how to do this is explained in the demo script itself).  The only necessary result of this is we now walk the user through creating some test data to manage with `alpenhorn` as part of the demo, rather than the demo starting with some data pre-existing.


ljgray (approved):
I think this looks fine

PR #287, opened 2025-03-31
feat(cli): Read file list via command-line option

### LINKED ISSUE #285: Ability to specify a list of files via text file to execute a CLI command on.
This is a normal things we do when dealing with files on cedar (i.e. have a list of files in a text file that we want to delete and or transfer).

Not sure if it would apply to everything in the `file` command group or a subset or comprise a separate command group with specific commands...

ketiltrout:
This adds a `--file-list` option to various commands to allow limiting the command to a list of ArchiveFiles specified in a text file.

The intention is that, say,
```
        alpenhorn node sync cedar_nearline cedar_online --file-list=<FILE>
```
should do the same thing as
```
        recall_filelist <FILE>
```
is currently doing on cedar.

Can also read from stdin (via `--file-list=-`), in which case `--check` mode will be automatically turned on if the user doesn't invoke `--force`. (Because the file-reading code will consume all the input, making it impossible for the user to interactively confirm the operation, in that case.)

The commands which accept this option are:

* `node clean`
* `node sync` (AKA `group sync`)
* `node verify`

Also: renamed the `--count` flag in `node verify` to `--check` to make it the same as all the other commands with a check/confirm/update flow. Not sure why it was different; the `node verify` tests seem to indicate I had been assuming this rename already.

Closes #285

### DIFF
--- a/alpenhorn/cli/cli.py
+++ b/alpenhorn/cli/cli.py
@@ -26,7 +26,7 @@ def check_then_update(
     1. If `do_check` is False, skips to step 6
     2. Calls `func` in "check" mode.
     3. If `do_update` is False, program exits
-    4. `click.confirm("Continue?")
+    4. Asks user for confirmation.
     5. If user declines confirmation, program exits
     6. If `do_update` is True, calls `func` in "update" mode.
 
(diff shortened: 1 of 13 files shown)


ljgray (approved):
Looks fine. 

I prefer `--file-list` to `--from-file`, but I think that both sort of have the issue where it isn't clear that, when this argument is omitted, all files are included. Something like `--restrict-files`? Although I don't love that.

ketiltrout:
Hmm.

ketiltrout:
I've changed it to `--file-list`.  In addition to being more clear, it avoids the clash with the `file create` flag.

> it isn't clear that, when this argument is omitted, all files are included

This is potentially a problem with a number of our optional arguments which restrict operation.  A lot of commands, for instance have an `--acq` option to limit a command to a specific acquisition.

ljgray:
Fair enough. I think that's fine

PR #296, opened 2025-04-07
fix(node stats): Handle large node totals

ketiltrout:
It appears that when (a) using a MySQL database and (b) the total size of a node is above some threshold, then the resulting total size is reported as a `decimal.Decimal` rather than a simple `int`.  In this case, the number can't be multiplied or divided by a `float`.

So, I've re-arranged this calculation and done an explicit cast to get around this.

Unfortunately, I can't seem to reproduce this using Sqlite, so I don't have a unit test for it at the momemt.  Perhaps an argument for doing a pass of the test suite with MySQL...


ljgray (approved):
That's annoying. I can't see any issue with this change, even if the error can't always be reproduced.

PR #299, opened 2025-04-12
refactor(pattern_importer): Don't extend alpenhorn tables

ketiltrout:
We're not going to remove support to allow extensions of the alpenhorn tables.  CHIME needs it.  (Also: peewee allows it; it would be difficult to prevent from within alpenhorn.) But, I think it's both unnecessarily confusing, and also not a demonstration of best practices, to do this in an example extension.

So, this PR replaces the extension of `ArchiveAcq` and `ArchiveFile` in the pattern exporter with a couple new tables (`AcqData` and `FileData`) to hold the extra metadata that it wants to store.

As a result of changing this I've re-ordered the steps in the demo to create the alpenhorn data index first and then initialise the extension second, which I think is a more intuitive way to do it.  I've also updated the end-to-end test of the daemon in `tests/daemon/test_entry.py` for this change.

### DIFF
--- a/demo/demo-script.md
+++ b/demo/demo-script.md
@@ -217,6 +217,20 @@ root@alpen1:/#
 
 Once at the root prompt, we can build the data index and start populating it.
 
+### Setting up the data index
+
+Creating the data index is simple, and can be accomplished with the alpenhorn CLI command:
+```
+  alpenhorn db init
+```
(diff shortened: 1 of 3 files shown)


jmaceachern (approved):
Great idea! Changes look good.

PR #301, opened 2025-04-14
docs: Move demo script into Alpenhorn docs

ketiltrout:
This has involved converting the document from MarkDown to ReStructuredText.  I've also make a short expansion to the script to indicate how to clone the repository and navigate to the `/demo/` directory, since the expectation is now that the script is going to be read on RTD rather than directly from the repo.

The version of the demo script here already contains the changes to the document made in both #299 and #300.

Plus various other doc tweaks.

This version of the documentation has been built as the hidden`doctest` version of the project on RTD and can be viewed here: https://alpenhorn.readthedocs.io/en/doctest/


jmaceachern (approved):
Demo looks great in RTD form.

PR #303, opened 2025-04-14
Mypy / ruff format fixes

ketiltrout:
## Errors found by `mypy`

`ruff`, explicitly, doesn't do type checking with type hinting, so I've experimented a bit with running `mypy` on the code to see what it would say.  The primary result is now I have Opinions on `mypy`, which this PR is too narrow to contain.  But has also found some problems that I've fixed here, which is great.  Still, I'm not convinced it's something we want running in our CI...

Most of these fixes are fixes to the type hints themselves, which is what I was expecting `mypy` to help me with, but there are a few actual (albeit minor) bugs which have also been fixed.

## ruff format

I also experimented in this branch with `ruff format` as a replacement for `black`.  It produces results that are very similar (by design: the `ruff` devs are overtly trying to mimic black), but it's not identical. The thing I liked most about `ruff format` is its ability to concatenate strings when the result would fit on a single line.

The code here has been run through `black` and then `ruff format` and then back through `black`, meaning there are a few formatting changes in the code that were changed by `ruff` from what `black` had done, but then were considered acceptable by the second `black` run, and not reverted. (Several other changes made by `ruff format` _were_ reverted.)

I think the formatting done by `ruff format` is good enough that, were we using `ruff` everywhere I'd suggest using it for formatting as well as linting, and ditch `black`, just to minimise the number of tools we have in play.  But, since we're not using `ruff` everywhere, I don't think it's worth switching.  We definitely shouldn't get ourselves into a situation where were using `black` in some projects and `ruff format` in others.

### DIFF
--- a/alpenhorn/cli/group/sync.py
+++ b/alpenhorn/cli/group/sync.py
@@ -35,7 +35,7 @@ def _run_cancel(
     ctx,
     group: StorageGroup,
     node: StorageNode,
-    acqs: list[ArchiveAcq],
+    acqs: set[ArchiveAcq],
     listed_files: set[ArchiveFile],
     show_acqs: bool,
     show_files: bool,
@@ -112,14 +112,14 @@ def _run_cancel(
     if first_time:
         if show_acqs or show_files:
             # Need to fetch the full request in this case
-            requests = ArchiveFileCopyRequest.select().where(
+            reqs = ArchiveFileCopyRequest.select().where(
                 ArchiveFileCopyRequest.id << afcrs
             )
 
         if show_acqs:
             acq_counts = {}
             acq_files = defaultdict(list)
-            for req in requests:
+            for req in reqs:
                 acq = req.file.acq
                 acq_counts[acq] = acq_counts.get(acq, 0) + 1
                 if show_files:
@@ -131,7 +131,7 @@ def _run_cancel(
                 if show_files:
                     echo("\n".join(sorted(acq_files[acq])))
         elif show_files:
-            paths = [str(req.file.path) for req in requests]
+            paths = [str(req.file.path) for req in reqs]
             echo("\n".join(sorted(paths)))
 
     # Do update
@@ -151,7 +151,7 @@ def _run_sync(
     ctx,
     group: StorageGroup,
     node: StorageNode,
-    acqs: list[ArchiveAcq],
+    acqs: set[ArchiveAcq],
     listed_files: set[ArchiveFile],
     show_acqs: bool,
     show_files: bool,
@@ -401,7 +401,7 @@ def run_query(
         try:
             group = StorageGroup.get(name=group_name)
         except pw.DoesNotExist:
-            raise click.ClickException("No such group: " + node_name)
+            raise click.ClickException("No such group: " + group_name)
 
     # Resolve acqs
     acqs = resolve_acq(acq)
--- a/alpenhorn/cli/options.py
+++ b/alpenhorn/cli/options.py
@@ -34,8 +34,7 @@ def cli_option(option: str, **extra_kwargs):
         args = ("--address",)
         kwargs = {
             "metavar": "ADDR",
-            "help": "Domain name or IP address to use for remote access to the "
-            "node.",
+            "help": "Domain name or IP address to use for remote access to the node.",
         }
     elif option == "all_":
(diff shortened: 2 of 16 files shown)


jmaceachern (approved):

alpenhorn/cli/group/sync.py
```diff
@@ -112,14 +112,14 @@ def _run_cancel(
     if first_time:
         if show_acqs or show_files:
             # Need to fetch the full request in this case
-            requests = ArchiveFileCopyRequest.select().where(
+            reqs = ArchiveFileCopyRequest.select().where(
```
ljgray, line 115: Was this unrelated to `mypy`, or did it want to change for some reason?

alpenhorn/cli/options.py
```diff
@@ -463,13 +462,13 @@ def files_in_nodes(
 
             # In this mode, there's no reason to continue once the set is empty
             if not node_files:
-                return []
+                return node_files
```
ljgray, line 465: Is it better practice to return `node_files` here, or just to `break` and have a single return statement?

alpenhorn/cli/options.py
```diff
@@ -531,7 +530,7 @@ def files_in_groups(
 
             # In this mode, there's no reason to continue once the set is empty
             if not group_files:
-                return []
+                return group_files
```
ljgray, line 533: Same question here

ljgray (approved):
Looks good. Curious to hear more about `mypy`

alpenhorn/cli/group/sync.py
ketiltrout: It was not really a required change, but it was something that was confusing `mypy`.

I was also using the `requests` variable name when printing results, eg.  https://github.com/radiocosmology/alpenhorn/blob/99b37026f0be51ad78beb709a2c1fcfe03753f1a/alpenhorn/cli/group/sync.py#L129

Really I made the change to prevent confusing our future selves that may not notice these two independent uses of the same variable name.

alpenhorn/cli/options.py
ketiltrout: Yeah, that's a good question that I don't know the answer two.  I can change it.

alpenhorn/cli/group/sync.py
ljgray: Ahh, makes sense. I agree that that makes it less confusing

ketiltrout:
> Curious to hear more about `mypy`

I'm mildly disappointed with its static analysis capabilities (maybe unjustly so).  For instance:

1) in `pull_async`, it notices that sometimes `ioresult` is `None` (fair), and then gets to the end of the function and complains here:

https://github.com/radiocosmology/alpenhorn/blob/99b37026f0be51ad78beb709a2c1fcfe03753f1a/alpenhorn/io/_default_asyncs.py#L195

that `None` has no `get` method.  That's true, but by that time in the function flow, `ioresult` can't be `None` anymore, something I would have liked `mypy` to figure out (it should have all the information needed to conclude that).  Not sure how it could really be fixed, other than disabling the check there with a `# type: ...` pragma.

2) In the CLI there are a number of multi-option processing functions like `resolve_node` which return a scalar when passed a scalar and a set when passed a list.

`mypy` is able to do what it calls "type narrowing" where it understands say, that the type of a variable is restricted when in a `if isinstance(var, type):` block.  That's great, but it seem to be completely bamboozled by assigning the conditional to a boolean:

https://github.com/radiocosmology/alpenhorn/blob/99b37026f0be51ad78beb709a2c1fcfe03753f1a/alpenhorn/cli/options.py#L309-L311

so it's still complaining that `node` might be scalar afterwards, which is a bit of a disappointment.

I don't think, even if `mypy` were to understand this construct, that it would grok the relationship between input type and return type for these functions.  If we wanted to use `mypy` more seriously we'd almost need to overload the prototypes here.  i.e.:

```(Python)
@overload
def resolve_node(node: str) -> StorageNode: ...
@overload
def resolve_node(node: list[str]) -> set[StorageNode]: ...
def resolve_node(node: str | list[str]) -> StorageNode | set[StorageNode]:
  # [ function body here ]
```

This kind of thing, and having to type all the variables/attributes makes me think we'd have a fair bit of work to do if we wanted to use `mypy` regularly for type checking.

PR #311, opened 2025-04-21
Dequeue tasks when stopping/reseting I/O for a node.

### LINKED ISSUE #309: lustrehsm: node re-init results in duplicated ready tasks
By "re-init" I mean `UpdateableNode.reinit` which is triggered by changing the `StorageNode`'s I/O class or I/O config (or sometimes when juggling node names around).

The duplication happens because the reinit results in a new instance of the I/O class being created, which doesn't have any memory of what the previous instance knew.  So, if there are deferred tasks in the mix (or really any long-running task), the new I/O instance won't consider them, and instead resubmits a bunch of new jobs for things already enqueued.

Possible solutions:
* empty out the FIFO of a reinited node (meaning all jobs need resubmission).  Not a big fan of this
* have the I/O instance scan its task FIFO on start up.  Would need some collaboration with the queue itself to do this.
* store information about running tasks outside of the I/O instance and/or have some sort of "persistent" memory that gets translated across reinit.  This might work, but likely only if the I/O class hasn't changed.  Unclear if this would be implemented on a class-specific basis, or globally by alpenhorn.

ketiltrout:
This is a four-piece (=commit) update to the daemon to deal with tasks which are left in the queue after the node/group they're targetting goes away.

## Piece 1: Allow setting fifo label value

That is, setting the value of the "fifo" label in the metrics generated by the queue.  This is preparation for the next piece.  The goal here is to avoid requiring the FIFO used in the queue (which can be anything hashable) to be something human-readable.

## Piece 2: Use a unique FIFO for each IO instance

With the freedom that piece one gives us, piece 2 changes the way FIFO keys used by I/O instances to submit tasks are generated.  Now a unique FIFO will be used by each instance of an I/O class.

FIFO keys started out as just StorageNode.name.  That worked until StorageGroups also started submitting tasks, and they were distinguished by group/node.  The change here increases that specialisation by guaranteeing that an I/O instance will always have a unique FIFO.

Calculation of the FIFO key has been pulled out of `alpenhorn.io.base` (where it really shouldn't have been) and instead is now generated by the UpdatableNode/Group when the I/O is instantiated, and passed to it.

The new `label_fifo()` call in the queue is used to give the key a useful name in the metrics.  The label is the same for all I/O isntances of a given Storage object, meaning they're (deliberately) merged together, obfuscating the fact that different instances of the Storage object's I/O framework actually have separate FIFOs.

*NB*: no real changes to the test suite here other than using actual `queue`s in places where I had been lazy and had substituted `None` in when the queue wasn't actually being used.

## Piece 3: Implement FIFO clearing
Adds a method, `clear_fifo`, to the queue which will dequeue all pending and deferred tasks in the specified FIFO.  After doing this, the FIFO may be optionally closed, preventing further putting new items into it.  Calling `put` on a closed FIFO will raise `KeyError`.

In-progress tasks are not affected, though if the FIFO they're running out of is closed, and these tasks requeue or defer themselves, they won't successfully be able to do that.

The `Task` has been updated to be able to handle this `KeyError` (by doing nothing other than printing a message that the item was discarded).  However, invoking `clear_fifo`, isn't done until Piece 4, so this situation will not happen.

*Also*: moved the `task_done` method without change to better organise the `queue` module.

## Piece 4: Dequeue tasks when stopping/reseting I/O for a node

When a node stops being updated (e.g. when deactivated), now pending tasks for the node are discarded from the queue (by calling `clear_fifo`).  This will happen both when the node/group is no longer available (because it was deactivated or the filesystem went away), but also as part of a re-init, since that's going to restart the I/O framework for the storage node/group.

This is effectively the same as killing the daemon and restarting it, (but only for this node), but it turns out that's a fairly safe thing to do, because we've done a good bit of work to make the daemon robust against that situation over the years. So, despite my original opinion on the option,  this closes #309 using the first of the proposed solutions.

I think this is the right way to deal with this because it also fixes another problem which I've been mulling over: how to deal with pending I/O when a node goes away.  And this seems like the right solution to that problem too: i.e. just throw it all away, and assume some future daemon will figure it out.

### DIFF
--- a/alpenhorn/daemon/auto_import.py
+++ b/alpenhorn/daemon/auto_import.py
@@ -422,7 +422,10 @@ def update_observer(
         # If this node isn't being auto-imported or we're force-stopping
         # it, delete a watcher if one was previously scheduled
         if node.name in _watchers:
-            _observers[node.io_class].unschedule(_watchers[node.name])
+            try:
+                _observers[node.io_class].unschedule(_watchers[node.name])
+            except KeyError:
+                pass  # wasn't being watched
             del _watchers[node.name]
     else:
         # If there's already a watcher for this node, do nothing
--- a/alpenhorn/daemon/update.py
+++ b/alpenhorn/daemon/update.py
@@ -178,8 +178,13 @@ class updateable_base:
             # Parse I/O config if present
             config = self._parse_io_config(storage.io_config)
 
+            # Generate a new fifo key and label it
+            fifo = (self.is_group, storage.name, time.monotonic())
+            label = f"Group {storage.name}" if self.is_group else f"Node {storage.name}"
+            self._queue.label_fifo(fifo, label=label)
+
             # Initialise I/O object
-            self.io = self.io_class(storage, config, self._queue)
+            self.io = self.io_class(storage, config, self._queue, fifo)
 
             return True
 
--- a/alpenhorn/daemon/update.py
+++ b/alpenhorn/daemon/update.py
@@ -46,6 +46,10 @@ class updateable_base:
     def __init__(self) -> None:
         raise NotImplementedError("updateable_base cannot be instantiated directly.")
 
+    def __del__(self) -> None:
+        # Stop updating
+        self.stop()
+
     @property
     def name(self) -> str:
         """The name of this instance."""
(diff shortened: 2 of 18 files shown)


ketiltrout:
We've been running this branch on cedar now, and I've noticed #312 as a potential improvement.

ljgray:
> We've been running this branch on cedar now, and I've noticed #312 as a potential improvement.

Nice. I got halfway through reviewing this yesterday, so I'll try to finish today.

PR #314, opened 2025-04-29
build: run build checks in GH workflows

ketiltrout:
The operative step here (other than simply checking that the build process completes successfully) is `twine check dist/*`, which checks that the metadata in the packages are acceptable to PyPI.

Also filled out the metadata in `pyproject.toml`.


ljgray (approved):
This looks good.

Since you're including an Issues link, it might be worth adding an issue template.

ketiltrout:
> it might be worth adding an issue template.

yep

jmaceachern (approved):

PR #316, opened 2025-04-30
doc: Update demo per Josh's notes

ketiltrout:
_Read this version of the demo on RTD here: https://alpenhorn.readthedocs.io/en/doctest/demo.html_

An unfortunately large fraction of the changes here are simply RST formatting fixes (and adding things like hint and tip callouts).

Important changes:
* I've rotated the background highlight for stuff occurring within the docker containers by 180 degrees in hue (meaning the background is a not-entirely-unpleasant purplish instead of yellow) with the hope that it makes it more clear where specific actions take place.
* I've added an extra container to the demo, which I've called `alpenshell` where the demo suggests you run the `alpenhorn` CLI commands (though it does point out you could use one of the storage hosts, too).  My reasoning is twofold:
  * it reinforces the idea that the CLI can be run anywhere that can access the database, and does not need to be run on a node containing a daemon or StorageNode
  * it prevents the awkwardness of having to start up a temporary version of `alpenhost1` to bootstrap the demo.
* the downside to the above is there's a potential for confusion when doing manual file creation on `alpenhost1`, since that can't be done on `alpenshell` (because it doesn't have the `demo_storage1` filesystem).  I've tried to guard against this by throwing in some `cd /data` commands into the `alpenhost1` command lists.  Should I give `alpenhost1` their own background colour?

Individual comments are addressed below.

### DIFF
--- a/doc/demo.rst
+++ b/doc/demo.rst
@@ -339,9 +339,9 @@ Creating the data index is simple, and can be accomplished by running the follow
 command with the ``alpenhorn`` CLI utility:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
-     alpenhorn db init
+   alpenhorn db init
 
 .. hint::
    Remember that all these ``alpenhorn`` commands need to be run inside the
@@ -351,7 +351,7 @@ On successful completion, the ``db init`` command will report the version of the
 database schema used to create the Data Index:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# alpenhorn db init
    Data Index version 2 initialised.
@@ -392,9 +392,9 @@ To initialise the database for the extension, run the ``demo_init``
 function provided by the extension:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
-     python -c 'import pattern_importer; pattern_importer.demo_init()'
+   python -c 'import pattern_importer; pattern_importer.demo_init()'
 
 If you get a ``ModuleNotFoundError: No module named 'pattern_importer'``
 error, you're probably not executing this command in the root-shell in
@@ -403,10 +403,10 @@ the ``alpenshell`` container.
 You should see a success message:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# python -c 'import pattern_importer; pattern_importer.demo_init()'
-   Plugin init complete complete.
+   Plugin init complete.
 
 Create the first StorageNode
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -422,14 +422,14 @@ that we'll create later).
 To create the group, which we'll call ``demo_storage1``, run:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
-     alpenhorn group create demo_storage1
+   alpenhorn group create demo_storage1
 
 This should create the group:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# alpenhorn group create demo_storage1
    Created storage group "demo_storage1".
@@ -446,9 +446,9 @@ StorageGroup contains only one StorageNode, the node and group have the
 same name, though that's not required.)
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
-     alpenhorn node create demo_storage1 --group=demo_storage1 --auto-import --root=/data --host=alpenhost1
+   alpenhorn node create demo_storage1 --group=demo_storage1 --auto-import --root=/data --host=alpenhost1
 
 This command will create a new StorageNode called ``demo_storage1`` and
 put it in the identically-named group. Auto-import (automatic monitoring for
@@ -456,7 +456,7 @@ new files) will be turned on; the mount point in the filesystem will be set
 to ``/data`` and we declare it to be available on host ``alpenhost1``:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# alpenhorn node create demo_storage1 --group=demo_storage1 --auto-import
                  --root=/data --host=alpenhost1
@@ -539,14 +539,14 @@ In the ``alpenshell`` container, at the root prompt, we can now activate the
 node:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    alpenhorn node activate demo_storage1
 
 Alpenhorn will acknowledge the command:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# alpenhorn node activate demo_storage1
    Storage node "demo_storage1" activated.
@@ -568,7 +568,7 @@ file that it can't find. But, generally, it's easier to get alpenhorn
 to initialise the node for us:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    alpenhorn node init demo_storage1
 
@@ -576,7 +576,7 @@ The initialisation is not performed by the alpenhorn CLI.  Instead the
 CLI will create a request in the database to initialise the node:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# alpenhorn node init demo_storage1
    Requested initialisation of Node "demo_storage1".
@@ -709,7 +709,7 @@ start a shell in the running container using ``docker exec``:
 Once in this root shell on ``alpenhost1``, we can create the first of our files:
 
 .. code:: console
-   :class: democontainer
+   :class: demonode1
 
    cd /data
    mkdir -p 2025/02/21
@@ -744,7 +744,7 @@ Once the file has been created, the lock file can be deleted, to trigger
 import of the file:
 
 .. code:: console
-   :class: democontainer
+   :class: demonode1
 
    rm -f 2025/02/21/.meta.txt.lock
 
@@ -773,7 +773,7 @@ You can use the alpenhorn CLI to see that this file is now present on
 the ``demo_storage1`` node:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# alpenhorn node stats
    Name             File Count    Total Size    % Full
@@ -808,7 +808,7 @@ appending, say, ``.temp`` to the name of the file we want to create.
 In the ``alpenhost1`` container:
 
 .. code:: console
-   :class: democontainer
+   :class: demonode1
 
    cd /data
    mkdir 2025/02/21/23
@@ -835,7 +835,7 @@ After file is fully written, it can be moved to the correct name. On
 most filesystems, this is an atomic operation:
 
 .. code:: console
-   :class: democontainer
+   :class: demonode1
 
    mv 2025/02/21/23/1324.dat.temp 2025/02/21/23/1324.dat
 
@@ -864,7 +864,7 @@ created in the database, because the ``ArchiveAcq`` record already exists:
 Now there are two files on the node:
 
 .. code:: console
-   :class: democontainer
+   :class: demoshell
 
    root@alpenshell:/# alpenhorn node stats
    Name             File Count    Total Size    % Full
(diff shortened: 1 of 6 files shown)


doc/demo.rst
```diff
@@ -54,22 +55,32 @@ installation by running the ``hello-world`` container:
...
 
-Once docker itself is running, you’ll also need to `install the docker
+.. hint::
+   If you get a permission error: "permission denied while trying to connect
+   to the Docker daemon socket", this means you lack the proper permissions to
+   run docker.  The best solution in this case is to have your sysadmin add you
+   to the ``docker`` group.  See `the Docker documenation on running docker as
+   a non-root user
+   <https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user>`__.
+
```
ketiltrout, line 65: > When running “docker compose up --detach alpdb”, got:
`unable to get image 'mysql:latest': permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.48/images/mysql:latest/json": dial unix /var/run/docker.sock: connect: permission denied`
^ Fixed by adding sudo. Should maybe address this in the demo or change commands throughout demo to have sudo?

I've added this hint to address this and point to the relevant Docker docs: better than running sudo is to join the `docker` group, which will give you access to the socket.

doc/demo.rst
```diff
@@ -144,27 +163,32 @@ latest MySQL image:
...
-   [+] Running 1/1
-    ✔ Container demo-alpdb-1  Started 
+   [+] Running 3/3
+    ✔ Network demo_default      Created                                         0.1s
+    ✔ Volume "demo_db_vol"      Created                                         0.0s
+    ✔ Container demo-alpendb-1  Started                                         2.6s
 
-You can use ``docker status`` or ``docker container ls`` to verify that
-the alpdb container is running:
+You can use ``docker stats`` or ``docker container ls`` to verify that the
```
ketiltrout, line 171: > -> After starting alpdb, `docker status` did not work for me, but `docker container ls` did.

Right.  It's `docker stats` not `status`

doc/demo.rst
```diff
@@ -210,108 +244,138 @@ Now we need to use some ``alpenhorn`` commands to create the Data Index
...
 running, and alpenhorn has been set up to use it.
 
+.. hint::
+   The reason alpenhorn is aware of the ``pattern_importer`` extension is
+   because it is listed as an extension to load in the alpenhorn config file,
+   which is available in the ``alpenshell`` at ``/etc/alpenhorn/alpenhorn.conf``.
+
+   You can also take a look at it on the docker host, in the ``/demo/``
+   subdirectory out of which you're running this demo.
+
```
ketiltrout, line 378: > -> Quick sentence to describe *how* the pattern_importer extension has been incorporated into the alpenhorn container image that we’re running and *how* alpenhorn has been set up to use it? Maybe point user to read demo_init function?

Here's a hint for that.  This is also reiterated below when the first daemon starts up and the user notices the `pattern_importer` extension being loaded.

doc/demo.rst
```diff
@@ -322,255 +386,263 @@ Fortunately, for the simple case in this demo, we don't have to worry about that
...
 
-     alpenhorn node create demo_storage1 --group=demo_storage1 --auto-import --root=/data --host=alpen1
+     alpenhorn node create demo_storage1 --group=demo_storage1 --auto-import --root=/data --host=alpenhost1
 
 This command will create a new StorageNode called ``demo_storage1`` and
-put it in the identically-named group. Auto-import will be turned on;
-the mount point in the filesystem will be set to ``/data`` and we
-declare it to be available on host ``alpen1``:
+put it in the identically-named group. Auto-import (automatic monitoring for
+new files) will be turned on; the mount point in the filesystem will be set
```
ketiltrout, line 450: > -> Auto import is mentioned for the first time but not explicitly explained (though it’s pretty intuitive what it means)

✅ Though this will also need to be addressed in the documentation proper.

doc/demo.rst
```diff
@@ -322,255 +386,263 @@ Fortunately, for the simple case in this demo, we don't have to worry about that
...
+   alpenhost1-1  | Feb 21 00:41:02 INFO >> [MainThread] Node demo_storage1: 46.77 GiB available.
+   alpenhost1-1  | Feb 21 00:41:02 INFO >> [MainThread] Updating node "demo_storage1".
+   alpenhost1-1  | Feb 21 00:41:02 INFO >> [MainThread] Updating group "demo_storage1".
+   alpenhost1-1  | Feb 21 00:41:02 INFO >> [MainThread] Main loop execution was 0.0s.
+   alpenhost1-1  | Feb 21 00:41:02 INFO >> [MainThread] Tasks: 1 queued, 0 deferred, 0 in-progress on 2 workers
+
+It will also run a job to see if there's anything needing clean-up on the
+node. This "tidy up" job helps the alpenhorn daemon recover from
+unexpected crashes by looking for and removing temporary files which the
+alpenhorn daemon may have not been able to clean up the last time it ran.
```
ketiltrout, line 626: > -> What does tidying up actually do?

✅

doc/demo.rst
```diff
@@ -687,73 +785,87 @@ Auto-importing files and temporary names
...
+   :class: democontainer
+
+   mv 2025/02/21/23/1324.dat.temp 2025/02/21/23/1324.dat
 
-   mv /data/2025/02/21/23/1324.dat.temp /data/2025/02/21/23/1324.dat
+.. hint::
+   By "atomic operation" we mean: on most filesystems there is never
+   a time during execution of the ``mv`` command when the destination
+   filename ``2025/02/21/23/1324.dat`` refers to a partial file.  Either
+   the destination file doesn't exist, or it exists and is complete.
```
ketiltrout, line 841: > -> Why do you mention the point about an atomic operation? In case the file is created before its contents are written in a `mv` command?

Definitely worth expanding on that term. Here's a hint.

doc/demo.rst
```diff
@@ -1075,66 +1216,74 @@ The alpenhorn daemon has the ability to transfer files between Storage
...
+   :class: democontainer
 
-   root@alpen1:/# alpenhorn file sync --from demo_storage1 --to demo_storage2 2025/02/21/meta.txt
+   root@alpenshell:/# alpenhorn file sync --from demo_storage1 --to demo_storage2 2025/02/21/meta.txt
    Request submitted.
 
-Transfers are always handled on the receiving side. After, perhaps, a
-short while, the daemon on ``alpen2`` will notice this request. First,
+Transfers are always handled on the receiving side (that is: by the daemon
+which considers the destination StorageGroup to be available). After, perhaps,
```
ketiltrout, line 1240: > -> Can requests for a certain node only be handled by a daemon running on that node or with access to the filesystem on the node? Is this related to `Daemon Host` field?

Perhaps sufficient...?

doc/demo.rst
```diff
@@ -1788,95 +1984,114 @@ transport node:
...
+   alpenhost3-1  | Mar 09 04:10:07 INFO >> [MainThread] Node "transport1" no longer available.
+   alpenhost3-1  | Mar 09 04:10:07 INFO >> [MainThread] Group "transport_group" no longer available.
 
-   alpen3-1  | Mar 09 04:10:07 INFO >> [MainThread] Node "transport1" no longer available.
-   alpen3-1  | Mar 09 04:10:07 INFO >> [MainThread] Group "transport_group" no longer available.
+.. hint::
+   A StorageGroup is only available to a daemon if at least one of its nodes is
+   available.  When we deactivate the ``transport1`` node, resulting it it no
+   longer being available, the ``transport_group`` also becomes unavailable because
+   there are no other active nodes on ``alpenhost3``.
```
ketiltrout, line 2013: > -> Does transport group become unavailable after transport1 is deactivated only because transport 1 is the only node in the group?

A good question.  Hopefully this hint answers it.

ketiltrout:
Other points:

> -> How is the root authentication happening/working in the demo?

Very deliberately being swept under the rug here.

The goal of the demo is neither to help users figure out how to do sysadmin their hosts nor to explain the details of how the docker multi-container set-up is working.  The user should learn that communication between nodes (for the purposes of file transfer) is via `rsync`, but we're not going to explain how to set-up unattended rsync between hosts here.  (Perhaps we will in the primary docs.)

Similarly, we're not explaining how we're authenticating with the database in the demo.

> -> What are other possible states in `alpenhorn file list` command? Corrupted? What is Suspect? Missing?

This is needs a link to the primary documentation, but that doesn't exist yet.

> -> What other “classes” of StorageNodes are available? I.e. clarify on this: “The "Transport" Group I/O class allows StorageNodes of any class to be added to the group, but requires all such nodes to have the "transport" storage type.” Archive and Transport are the only types?

Eventually, when such documentation exists, we'll need a pointer in the demo here to the list of in-built I/O classes.  The demo really doesn't explain what an I/O class is at all.

jmaceachern:
> The goal of the demo is neither to help users figure out how to do sysadmin their hosts nor to explain the details of how the docker multi-container set-up is working. The user should learn that communication between nodes (for the purposes of file transfer) is via rsync, but we're not going to explain how to set-up unattended rsync between hosts here. (Perhaps we will in the primary docs.)

That makes total sense^.

doc/demo.rst
```diff
@@ -210,108 +244,138 @@ Now we need to use some ``alpenhorn`` commands to create the Data Index
 infrastructure in it. The data index must exist before we can start the
 first alpenhorn daemon.
 
-To create the data index we’ll need access to the database. We’ll do
-that by creating a temporary container running the image from the first
-alpenhorn node (``alpen1``):
+To create the data index we'll need access to the MySQL database housing it.
+This can't be done from the docker host, so we'll create a separate docker
+container (called ``alpenshell``) which we'll use for the duration of this
+demo to interact with alpenhorn.
```
jmaceachern, line 250: Can you remind me again why this can't be done from the docker host? Is it because the daemon will start on alpenhost1 and will get mad because the data index isn't created yet?

I think alpenshell is a super elegant solution for this. I wonder if it's worth removing the part about "this can't be done from the docker host". E.g.:

```
To create the data index we'll need access to the MySQL database housing it.
We'll make a docker container (called ``alpenshell``) which we'll use to create
the data index and to interact with alpenhorn for the remainder of the demo.
```

Of course, I also understand if you want the reader to know about this subtlety with creating the data index.

doc/demo.rst
```diff
@@ -1788,95 +1984,114 @@ transport node:
...
-device from the machine
+If this were a real transport device, the next steps would be to:
 
-Next, the trasnport device would need to travel (via Sneakernet) from
-the site containing ``alpen3`` to the site containing ``alpen1`` where
+- unmount the filesystem
+- eject the media
+- remove the physical storage device from the machine
+
+After this, the trasnport device would need to travel (via Sneakernet) from
```
jmaceachern, line 2021: Small typo: `trasnport` -> `transport`

jmaceachern (approved):
Looks great. The demo was clear before, but all of these changes have improved the clarity even more. 

As for giving alpenhost1 its own colour, I think that would make it even more clear what commands should be issue on which host. However, in many of the text boxes in the RTD showing issued commands, you often include the terminal prompt: e.g. `root@alpenhost1:/data# ....` which I think makes it clear.

I'll leave it up to you to whether or not to give the alpenhosts their own colours, but I think its pretty clear. However, I'll happily review if you do decide to go ahead with that.

doc/demo.rst
ketiltrout: > Can you remind me again why this can't be done from the docker host?

Primarily, it's because the docker host doesn't have access to the MySQL port (which is only available on the multi-container virtual network).

But also: there's no reason to assume there's a working alpenhorn installation on the docker host.

> I wonder if it's worth removing the part about "this can't be done from the docker host"

Maybe.  I was trying to motivate the need for a shell in the container.  Do you think that's not needed?

ketiltrout:
> However, in many of the text boxes in the RTD showing issued commands, you often include the terminal prompt: e.g. root@alpenhost1:/data# .... which I think makes it clear.

I show the prompt when showing command output.  But when presenting a command, I don't want to have the prompt in the highlight box because I want readers to be able to easily drag-select everything in the box so they can then paste the comamnds into their terminal.

doc/demo.rst
jmaceachern: That makes sense. Thanks for explaining again.

And good point about motivating the use of the shell. I like that reasoning.

PR #320, opened 2025-05-09
feat(metrics): Add "daemon" label to metrics

ketiltrout:
Forces a "daemon" label to be bound whenever creating a metric.  The default value of this metric is the hostname we're running on (specifically, the value returned by `util.get_hostaname()`, which may be a value specified in the config file).

This is prep work for relaying metrics from cedar through the database.

Note: I've deleted the `test_unlabelled_metric` test because, with this change, alpenhorn can no longer create unlabelled metrics, meaning the test can't be performed anymore. (Support for unabelled metrics hasn't been removed from the the `metric` module).

### DIFF
--- a/alpenhorn/common/metrics.py
+++ b/alpenhorn/common/metrics.py
@@ -90,18 +90,18 @@ class Metric:
         global _metrics
 
         # Bind the "daemon" label if not already bound
-        if "daemon" not in bound:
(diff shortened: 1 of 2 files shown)


ljgray (approved):
Seems fine, except for the failing test

PR #321, opened 2025-06-05
refactor(config): Access config via `common.config.get`

ketiltrout:
Instead of directly accessing the `config.config` dict, this introduces a function, `common.config.get`, as the standard way off accessing the config.

The primary benefit here is that this centralises handling of missing config parameters (either by returning a default or by raising an exception) and also provides a convenient place for type checking/co-ercion of config values.

There are also specific functions for numerical types: `get_int`, `get_float` which can do bounds checking.  Also `get_bytes` which handles numbers specified with "k/M/G" suffixes (and used to be the bespoke `common.logger._max_bytes_from_config` function).

The config dict itself is renamed to `config._config` to indicate it shouldn't be accessed directly.

With this change to how the alpenhorn config is accessed, there is now a convenient way of specifying default values at the place a config parameter is used. So, I've deleted the `config._default_config` dict, replacing it with defaults in the `config.get` calls instead, which I think will aid with program understanding (because the default value is now in the place where it's used, rather than in some other module somewhere).

A side effect of this change is that all config errors are now reported to the user with a consistent `click.ClickException`, instead of the mess of exceptions previously seen.  Also better handling of incorrect configuration parameters types.

### DIFF
--- a/alpenhorn/common/config.py
+++ b/alpenhorn/common/config.py
@@ -98,11 +98,11 @@ Example config:
             # Maximum number of rotated files to keep.  Must be at least one
             # Rotated files append an integer to the name specified (e.g.
             # "file.log.1", "file.log.2" etc.).
-            backup_count: 100
+            backup_count: 10
 
             # Size, in bytes, at which log file rotation occurs.  May include a suffix:
             # k, M, or G.
-            max_bytes: 4G
+            max_bytes: 4M
 
 
     # Configure the operation of the local daemon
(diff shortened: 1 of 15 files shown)


alpenhorn/common/config.py
```diff
@@ -98,11 +98,11 @@
             # Maximum number of rotated files to keep.  Must be at least one
             # Rotated files append an integer to the name specified (e.g.
             # "file.log.1", "file.log.2" etc.).
-            backup_count: 100
+            backup_count: 10
```
ljgray, line 101: What's the reason for changing this (and `max_bytes`)?

ljgray (approved):
This definitely looks cleaner

PR #325, opened 2025-07-22
fix(update_group): Only run pre-pull search if needed

ketiltrout:
I think I over-corrected in #190 when I created a group-level task to search for existing unregistered files before a pull happens.  It was, in theory, the correct thing to do, but in practice, it leaves us with a mountain of mostly unnecessary tasks in the queue.

In this PR, I've changed this to make the group-level pre-pull search optional.  The idea being: in cases where the destination node for a group pull is deterministic, there's no need to search the whole group, and the search can be relegated to the target node itself.

In practice, this means that we can get rid of the group-level search from all group classes that take only a single node (e.g. `Default` and `LustreQuota`) as well as from `LustreHSM` (because there the destination node is determined solely by file size).

Among the in-built I/O types, this leaves only the `Transport` group where a full, group-level search needs to happen.  Whether or not the search happened is now passed on to the node-level pull, which will do the search itself before running the pull if it hasn't happened.

In the Group I/O layer, this renames the `pull_force` method to `pull` and the `pull` method to `pull_search`, and introduces a Group I/O class attribute, `do_pull_search` to specify which classes should do the full search.

I've left the default implementation of `pull_search` in `DefaultGroupIO` even though that class sets `do_pull_search` to False, meaning that it will never be used, but it's like that to make it easier for subclasses to implement the behaviour when needed.  I've also pulled the decision making over whether to use group-level `pull` or `pull_search` out to the `UpdateableGroup`, which is probably where it should always have been.

The tests for `pull_search` have been moved from `test_defaultgroup` into `test_transport` because the DefaultGroup can no longer perform the tests.

### DIFF
--- a/alpenhorn/io/_default_asyncs.py
+++ b/alpenhorn/io/_default_asyncs.py
@@ -15,7 +15,13 @@ from ..common import metrics
 from ..common.metrics import Metric
 from ..common.util import timeout_call
 from ..daemon.update import RemoteNode
-from ..db import ArchiveFileCopy, ArchiveFileCopyRequest, utcnow
+from ..db import (
+    ArchiveFile,
+    ArchiveFileCopy,
+    ArchiveFileCopyRequest,
+    StorageNode,
+    utcnow,
+)
 from . import ioutil
 
 if TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


alpenhorn/io/_default_asyncs.py
```diff
@@ -407,38 +483,10 @@ def group_search_async(
...
-            ).where(
-                ArchiveFileCopy.file == req.file,
-                ArchiveFileCopy.node == node.db,
-            ).execute()
+        _force_check_filecopy(req.file, node.db, node.io)
         return
 
-    # Otherwise, escallate to groupio.pull_force to actually perform the pull
-    groupio.pull_force(req)
+    # Otherwise, escallate to groupio.pull to actually perform the pull
```
ljgray, line 491:
```suggestion
    # Otherwise, escalate to groupio.pull to actually perform the pull
```

ljgray (approved):

PR #332, opened 2025-09-09
refactor: Move io and scheduler modules under daemon

ketiltrout:
When I first subdivided the code, there was still the idea that the client was going to support doing I/O, which meant it was going to need access to both the I/O framework and the scheduler.

But since the CLI refactor removed all I/O from it, that's no logner the case.  So, let's move the `scheduler` and `io` modules under `daemon`, since that's the only part that needs them now.

(The only part of this change I don't like is all the top-level `alpenhorn` submodules now only begin either with a "c" or a "d", which seems ... odd.)

This and #333 overlap to a degree.  Probably one or the other should be reviewed and merged first before the other is reviewed, though the order I don't think matters.


ketiltrout:
There are a lot of problems with the docstrings.  I have another branch (doctest) where I'm slowly trying to clean them up.

ketiltrout:
Linting fixes are in #334

PR #333, opened 2025-09-10
refactor: Drop the use of `typing.TYPE_CHECKING`

ketiltrout:
Given we're not using a static type checker in our CI, the imports protected by `TYPE_CHECKING` have gotten too out-of-date to be useful. So, let's try getting rid of `TYPE_CHECKING`, and deal with circular import some other way.

This also words around sphinx autodoc problems associated with the use of `TYPE_CHECKING`.

Other than monkeying about with imports, the only other change I've made here is to re-import `UpdatedableNode/Group` and `RemoteNode` up a level in `alpenhorn.daemon` to make importing these easier.  The pattern_importer extension demonstrates that third-parties are going to want to have access to these.

This and #332 overlap to a degree.  Probably one or the other should be reviewed and merged first before the other is reviewed, though the order I don't think matters.

### DIFF
--- a/alpenhorn/cli/cli.py
+++ b/alpenhorn/cli/cli.py
@@ -2,16 +2,12 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
+from collections.abc import Callable
 
 import click
 
 from ..common.logger import echo
 from ..db import connect, schema_version
-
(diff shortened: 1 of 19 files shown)


ketiltrout:
Linting fixes are in #334

PR #337, opened 2025-09-16
Disperse DB-modifying `pullutil` remnant to the DB module

ketiltrout:
This is the second and last part of the I/O framework refactor (#336), which is concerned with moving the DB-modifying `ioutil` remnant that I had dumped into `daemon.pullutil` in the last PR to more appropriate locations under `alpenhorn.db`.  Specifically:
* `post_add` moves to `ArchiveFileCopy.trigger_autoactions`, attached to the newly-created file copy record. 
* `copy_request_done` moves to `ArchiveFileCopyRequest.finish`.

The changes needed to accomplish this aren't that great, but do result in a bit of a mess, so I've made four logical commits, if that makes reviewing this easier:

## 1. Move `post_add` to `ArchiveFileCopy.trigger_autoactions`

The `trigger_autoactions` doesn't really do anything with the file copy record it's attached to and, as a result, it might be argued it shouldn't be in this class, but it seems the most natural: the former `post_add` needed the `StorageNode` of the new file copy and the `ArchiveFile` of the new copy, and those naturally combine to produce an `ArchiveFileCopy`, so it seems most natural.

Because it doesn't actual interact with its `ArchiveFileCopy` instance, other than reading the node and file elements from it, the implementation of the method is essentially unchanged.

## 2. Split the Node I/O method `filesize` into two parts: `filesize` and `storage_used`

The `filesize` method had an optional `bool` parameter which selected the size of the file vs. the size on disk, which is
what's being broken up here.  The motivation here is really so I can use `storage_used` in the next commit, but I think this removes some unnecessary confusion about what "actual size" means, so I think this is a good change, even without that.  Another distinction: `filesize` may never return `None`, but `storage_used` may.

This change simplifies the `filesize` implementation in LustreHSM so that it only ever returns None, which saves us an otherwise unnecessary `stat` on nearline, which is a nice side-benefit.

## 3. Replace the Node I/O `io` parameter in `copy_request_done` with a `node_to` and `size` parameters.

The goal here is to remove the explicit dependence on the I/O framework from the `copy_request_done` function.  It used
to take the Node I/O instance and extract from that two things: the underlying `StorageNode` and the size-on-disk of the
file.  Now those two things are passed in directly, instead of passing in the Node I/O instance itself.

The `size` parameter can either be an explicit size, or a function (i.e. the new `storage_used` method) that provides the size.  In practice, the only callers to this function use the callable option, but I don't see a reason not to have the other option available.

## 4.  Move `copy_request_done` to `ArchiveFileCopyRequest.finish`

I think this is a very natural (and simple) move, which probably should have been where it always was implemented, but I think I was less willing to have extra methods in the table models back then.  Changes to the implementation is very light, the only thing of substance is modifying the `ArchiveFileCopyRequest` instance in-place to mark it complete, rather than using an UPDATE query.


ketiltrout:
If someone can come up with a better name for the method than `storage_used`, I'm all ears.

ljgray (approved):

PR #344, opened 2025-09-29
feat: Extension API

ketiltrout:
This change create the public Extension API in `alpenhorn.extensions`
which is used by third-parties to create Alpenhorn extensions.

Extension _modules_ haven't changed much: they're still listed in the
"extensions" section of the alpenhorn config and still must provide
a `register_extensions` function.  However, instead of the dict that
this function used to return, now it just returns a list of `Extension`
instances.

Instantiating classes makes it easier to be flexible with different
kinds of extensions, while collecting the common extension code in
one place.

Effectively, I suppose, this changes what is meant by the term
"Alpenhorn Extension".  Previously, and extension was a _module_
which provided certain functionality (described by the dict returned
by `register_extensions`).  Now extensions are class instances providing
that functionality, and the contents of the dict are now passed to the
class initialisers.  To reflect this (and to prevent old-style extension
modules from getting loaded), the extension hook has been pluralised to
`register_extensions`.

Operationally, this change has meant extension loading has been broken
up in to two parts: `find_extensions`, responsible for loading the
listed extension modules and collecting the Extension instances they
provide, and then an `init_extensions` (run in several parts) for
actually activating the extensions in alpenhorn.

Requires (and contains) all of #340 #342 #343 which should probably be reviewed first.

Breaks interoperability with:
* `chimedb.core`
* `alpenhorn-chime`


ketiltrout:
Tests are failing because they need: https://github.com/chime-experiment/chimedb/pull/54

alpenhorn/extensions/import_detect.py
```diff
@@ -0,0 +1,114 @@
...
+    a callable object providing a detection routine which will be called
+    when importing new files to determine if the file being considered is
+    a valid data file.  It will be passed a two positional parameters:
+
+      * `path`: a `pathlib.PurePath` giving the path relative to the node
+        root to the file being imported.
+      * `node`: a `UpdateableNode` instance of the node on which we're
+        importing the file.
+
+    The funciton should return a two-tuple.  If the detection fails, this
```
ljgray:
```suggestion
    The function should return a two-tuple.  If the detection fails, this
```

